PR #24, opened 2022-11-08
Feature/active account

MagoDelBlocco:
Use slightly modified readToken method to retrieve logged in account name

Changed readToken method to accept a boolean parameter which specifies whether to return the password or account name, and used it in the `account` function.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -274,7 +274,7 @@ program
       const authToken = await readToken(true).catch(() => undefined);
 
       if (!authToken) {
-        console.log("Unauthorized. You are not logged in.");
+        console.log("You are not logged in. Run 'genezio login' before displaying account information.");
       } else {
         console.log("Logged in as: " + authToken);
       }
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -271,20 +271,13 @@ program
...
-            console.log("Unauthorized. You are not logged in.")
-          }
-        })
-        .catch(() => {
-          console.log("Cannot access keychain.")
-        })
+      const authToken = await readToken(true).catch(() => undefined);
+
+      if (!authToken) {
+        console.log("Unauthorized. You are not logged in.");
```
andreia-oca, line 277:
```suggestion
        console.log("You are not logged in. Run 'genezio login' before displaying account information.");
```

src/index.ts
```diff
@@ -271,20 +271,13 @@ program
   .description("Display currently logged in account.")
```
andreia-oca, line 271:
```suggestion
  .description("Display information about the current account.")
```

andreia-oca (changes requested):

PR #42, opened 2022-11-17
Project ID can now be found in classesInfo

Virgil993:
and is used to form the url

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -22,6 +22,7 @@ import { NodeJsBundler } from "./bundlers/javascript/nodeJsBundler";
 import { NodeJsBinaryDependenciesBundler } from "./bundlers/javascript/nodeJsBinaryDepenciesBundler";
 import { getProjectConfiguration } from "./utils/configuration";
 import { REACT_APP_BASE_URL } from "./variables";
+import log from "loglevel";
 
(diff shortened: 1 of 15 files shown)


src/commands.ts
```diff
@@ -21,6 +21,7 @@ import {
 import { NodeJsBundler } from "./bundlers/javascript/nodeJsBundler";
 import { NodeJsBinaryDependenciesBundler } from "./bundlers/javascript/nodeJsBinaryDepenciesBundler";
 import { getProjectConfiguration } from "./utils/configuration";
+import { REACT_APP_BASE_URL } from "./variables";
```
radu1122, line 24: i think this is not used

radu1122 (approved):
looks good

src/commands.ts
```diff
@@ -239,6 +242,9 @@ export async function deployClasses() {
   await generateSdks(functionUrlForFilePath);
 
   reportSuccess(classesInfo, configuration);
+  
+  let projectId = classesInfo[0].projectId
```
vladiulianbogdan, line 246: Add this in the `reportSucess` method.

radu1122: it can't be added, the method its also used for local env

radu1122 (approved):

PR #45, opened 2022-11-21
Handled both errors by adding a specific catch for that  error in gen…

Virgil993:
…erateSdk and in DeployClasses

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -274,7 +274,9 @@ export async function deployClasses() {
 
   // wait for all promises to finish
   await Promise.all(promisesDeploy);
-  await generateSdks(functionUrlForFilePath)
+  await generateSdks(functionUrlForFilePath).catch((error)=>{
+    throw error
+  })
 
   reportSuccess(classesInfo, configuration);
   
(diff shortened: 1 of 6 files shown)


src/commands.ts
```diff
@@ -4,6 +4,7 @@ import { deployClass } from "./requests/deployCode";
 import generateSdk from "./requests/generateSdk";
 import listProjects from "./requests/listProjects";
 import deleteProject from "./requests/deleteProject";
+import axios, { AxiosError } from "axios";
```
radu1122, line 7: i think this is not used

src/variables.ts
```diff
@@ -1,4 +1,10 @@
+// export const REACT_APP_BASE_URL = "https://dev.app.genez.io";
+// export const BACKEND_ENDPOINT = "http://localhost:8080";
+// export const GENERATE_SDK_API_URL = "https://dev-sdk-api.genez.io";
+
 export const REACT_APP_BASE_URL = "https://app.genez.io";
 export const BACKEND_ENDPOINT = "https://api.genez.io";
```
radu1122, line 6: use dev variables please

PR #53, opened 2022-11-30
Improve the bundle time during genezio local command.

vladiulianbogdan:


src/bundlers/javascript/nodeJsBundler.ts
```diff
@@ -3,238 +3,181 @@ import path from "path";
...
-        }
+    async #bundleJavascriptCode(
+        filePath: string,
+        tempFolderPath: string,
+        mode: "development"|"production"
+    ): Promise<void> {
+        const outputFile = `module.js`;
+
+        // eslint-disable-next-line no-async-promise-executor
+        await bundle("./" + filePath, "development", [webpackNodeExternals()], {
```
radu1122, line 110: I think we should call it with mode

PR #55, opened 2022-12-01
Show webpack errors.

vladiulianbogdan:


src/utils/webpack.ts
```diff
@@ -39,6 +40,9 @@ export function bundle(
             }
 
             if (stats?.hasErrors()) {
+                stats?.compilation.getErrors().forEach((error) => {
+                    log.error(error.message)
```
radu1122, line 44: I think it should be merged with https://github.com/Genez-io/genezio/pull/48

PR #61, opened 2022-12-13
Properly handle multiple envs in CLI.

vladiulianbogdan:
Replace NODE_ENV from variables.ts with actuals values when building.


vladiulianbogdan:
If you want to install locally a dev version of the genezio CLI you have to run one of the following commands:

```
npm run install-locally-dev
npm run build-dev && npm install -g
```

If you want to install locally a production version of the genezio CLI you have to run one of the following commands:

```
npm run install-locally
npm run build && npm install -g
```

PR #63, opened 2022-12-19
GNZ-253 Improv/refactor the examples

vladiulianbogdan:


examples/blockchain/client/src/sdk/blockchainServer.sdk.js
```diff
@@ -0,0 +1,13 @@
+import { Remote } from "./remote.js"
```
radu1122, line 1: should we ad the SDK generated file in the repo?

andreia-oca (commented):
I think that `/tests/integration/test_blockchain.py` and `/tests/integration/test_todo_list.py` are not working right now because they've been moved, but not changed to reflect that.

andreia-oca (approved):

PR #64, opened 2022-12-22
added multi region support

radu1122:


examples/todo-list/frontend/src/backend-sdk/remote.js
```diff
@@ -0,0 +1,39 @@
+/**
```
vladiulianbogdan, line 1: I think this should be removed. Isn't it already in .gitignore? If not, can you please add it?

src/models/projectConfiguration.ts
```diff
@@ -19,177 +20,237 @@ export enum Language {
...
+        "The configuration file should contain at least one class."
+      );
     }
 
-    const classes = await Promise.all(unparsedClasses.map((c) => ClassConfiguration.create(c)))
+    const classes = await Promise.all(
+      unparsedClasses.map((c) => ClassConfiguration.create(c))
+    );
+
+    if (configurationFileContent.region) {
```
vladiulianbogdan, line 217: I think I would put this check inside in the `ProjectConfiguration` because we have there all the verifications.

vladiulianbogdan (approved):

PR #68, opened 2022-12-29
GNZ-326 Add better sanitization for `genezio.yaml`

andreia-oca:


src/models/projectConfiguration.ts
```diff
@@ -67,12 +76,68 @@ export class MethodConfiguration {
...
+    }
+
     return new MethodConfiguration(
       methodConfigurationYaml.name,
       type,
       methodConfigurationYaml.cronString
     );
   }
+
+  static isCronStringValid(unparsedCronString : string): boolean {
```
radu1122, line 91: I would also add a check for the case * * * * * *

radu1122 (approved):

PR #74, opened 2023-01-10
GNZ-365 Fix error catching for `genezio local`

andreia-oca:
Fix the following bugs for `genezio local`:
* gracefully exit if `genezio.yaml` is not found
* gracefully catch error if port is in used


radu1122 (approved):
great work

vladiulianbogdan (approved):

PR #77, opened 2023-01-11
Changes to local environment such that the request is parsed in the s…

vladiulianbogdan:
…ame way as AWS API Gateway.

### DIFF
--- a/src/utils/genezioRequestParser.ts
+++ b/src/utils/genezioRequestParser.ts
@@ -1,0 +1,45 @@
+/**
+ * Check if the body is of type binary.
+ * @param contentType The content type of the request
+ * @returns True if the body is binary, false otherwise.
+ */
+function bodyIsBinary(contentType: string) {
+    if (!contentType) {
+        return true
+    }
+
+    const components = contentType.split("/")
+    // TODO check if components length is 2.
+    const [mimeType, subType] = components
+    console.log(mimeType, subType)
+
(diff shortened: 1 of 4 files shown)


examples/webhook/client/test-webhook-sdk.js
```diff
@@ -1,5 +1,30 @@
-import { HelloWorldCronExample } from "./sdk/index.sdk.js"
+import axios from 'axios';
```
radu1122, line 1: this should be in gitignore

src/utils/genezioRequestParser.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param contentType The content type of the request
+ * @returns True if the body is binary, false otherwise.
+ */
+function bodyIsBinary(contentType: string) {
+    if (!contentType) {
+        return true
+    }
+
+    const components = contentType.split("/")
+    // TODO check if components length is 2.
```
radu1122, line 12: i think this check should be added

src/utils/lambdaHander.ts
```diff
@@ -22,25 +22,51 @@ exports.handler =  async function(event, context) {
         }
         return;
```
radu1122, line 23: can you also add the check for object[method] is not a function for cron?

radu1122 (approved):

PR #88, opened 2023-01-20
Feature/gnz 420 ast to backend

radu1122:


src/commands.ts
```diff
@@ -323,7 +356,7 @@ export function reportSuccess(
   }
 }
 
-export async function generateSdks(urlMap: any) {
+export async function generateSdks(urlMap: any): Promise<any> {
```
vladiulianbogdan, line 359: Replace `Promise<any>` with actual type.

vladiulianbogdan (approved):

PR #89, opened 2023-01-24
Feature/Deploy all classes in one request

vladiulianbogdan:


examples/hello-world/client/test-hello-sdk.js
```diff
@@ -1,4 +1,4 @@
-import { HelloWorldService } from "./sdk/hello.sdk.js"
```
radu1122, line 1: this should be in gitignore

PR #92, opened 2023-01-27
Send AST during project deployment.

vladiulianbogdan:


radu1122 (approved):
i don't see where you send the AST to the backend

PR #108, opened 2023-02-14
GNZ-416: Message to update CLI if version is not supported anymore.

vladiulianbogdan:
Send the CLI version in HTTP request headers to the backend and react when the backend says that the version is not supported.


radu1122 (approved):
Nice

PR #117, opened 2023-02-24
Add alternative login

MagoDelBlocco:
Adds an alternative login method that uses puppeteer. This was done to make genez.io compatible with WSL environments.

Typical usage: `genezio login --alt` -- will start the login process in a new chromium window. If `PUPPETEER_DATA_DIR` is not set, the user will need to login to Google/GitHub, then to app.genezio

To login headlessly: `genezio login --alt --headless`. The user should have `PUPPETEER_DATA_DIR` set, and has to be logged in to Google and genezio beforehand.


andreia-oca:
This PR is obsolete. We implemented a non-interactive way to login with personal access tokens.

The new approach works for CI/CD environments, windows, macos, linux and WSL.

I will close this for now.

andreia-oca:
This PR is obsolete. We implemented a non-interactive way to login with personal access tokens.

The new approach works for CI/CD environments, windows, macos, linux and WSL.

I will close this for now.

PR #145, opened 2023-03-15
Convert non-standard crons to standard crons for local deployment

MagoDelBlocco:
`2/5 * * * *` is a non-standard cron string as per https://crontab.guru/#2/5_*_*_*_* . The `node-cron` library doesn't support this format, although both Windows, Linux and AWS Lambda support it. The easiest solution was to detect these strings when deploying locally, and convert them to standard cron strings that have the same meaning:

`2/5 * * * *` == `2-59/5`

Rinse and repeat for all fields.

### DIFF
--- a/src/localEnvironment.ts
+++ b/src/localEnvironment.ts
@@ -188,6 +188,8 @@ export async function stopCronJobs(cronHandlers: LocalEnvCronHandler[]) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


andreia-oca (changes requested):
Please add unit tests for this change.

andreia-oca (approved):
LGTM

PR #146, opened 2023-03-15
added python support for sdk

radu1122:


src/utils/sdk.ts
```diff
@@ -32,12 +32,16 @@ export async function writeSdkToDisk(sdk: GenerateSdkResponse, language: Languag
         debugLogger.debug("No SDK classes found...")
         return 
     }
+    let extension: string = language as string;
```
vladiulianbogdan, line 35: handle all languages in the same way

PR #148, opened 2023-03-16
Fix request hanging on error issue. Return proper JSON RPC response w…

vladiulianbogdan:
…hen request is malformed.


radu1122 (approved):
It's ok

PR #160, opened 2023-03-20
Gnz 595 keep info after refresh

stefandarius:
Added field `name` in `/get-ast-summary` request to the local server, so the Test Interface can identify local projects and match them to saved state of the Interface for that project


radu1122 (approved):
all good

PR #166, opened 2023-03-24
faster ts time on genezio local

radu1122:


src/bundlers/typescript/nodeTsBundler.ts
```diff
@@ -221,36 +221,37 @@ export class NodeTsBundler implements BundlerInterface {
...
-                            ":" +
-                            error.loc?.split(":")[0]
-                        );
-                    } else {
-                        log.info(
-                            "file: " + error.moduleIdentifier + ":" + error.loc?.split(":")[0]
-                        );
-                    }
+                log.error(error);
+                // if (error.details?.includes("ts-loader-default")) {
```
vladiulianbogdan, line 225: How do we log errors now?

radu1122: fixed

PR #202, opened 2023-04-13
optional params in structs + return object instead of dicts python

stefandarius:


src/generateSdk/sdkGenerator/SwiftSdkGenerator.ts
```diff
@@ -80,6 +78,7 @@ class SdkGenerator implements SdkGeneratorInterface {
     };
 
     for (const classInfo of sdkGeneratorInput.classesInfo) {
+      // const externalTypes: Node[] = [];
```
vladiulianbogdan, line 81: remove this

PR #285, opened 2023-06-02
GNZ-817 Auto-delete `/tmp` folders for dart projects and `genezio local`

costinsin:
# Auto-delete `/tmp` folders for dart projects and genezio local

## Type of change

- [x] 🐛 Bug Fix

## Description
Previously, there were a lot of temporary folders created if dart projects were deployed using `genezio deploy` or if any kind of project was tested with `genezio local`.

Fixed this by removing all those folders in case of errors.

Also, updated the name of the temporary folders to start with `genezio-` prefix, instead of `foo-`.

## Checklist

- [ ] I have added tests;

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "genezio",
-  "version": "0.3.12",
+  "version": "0.4.0",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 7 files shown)


radu1122 (approved):
looks good

PR #309, opened 2023-06-19
Change webpack bundler with esbuild

costinsin:
# Pull Request Template

## Type of change

- [x] 🧑‍💻 Improvement

## Description
We made this change for multiple reasons:
- Faster build times
- Esbuild can correctly identify all required dependencies, so we can copy only the required ones when we deploy a project, lowering the uploaded archive file size.
- Esbuild knows how to bundle ESModules. This is useful because there are specific libraries out there that work only with ESModule `import` statement.

## Checklist


src/bundlers/node/nodeJsBundler.ts
```diff
@@ -0,0 +1,336 @@
...
+
+      dependencyMap.set(dependencyName, dependencyPath);
+    });
+
+    bundlerInput.extra.dependenciesInfo = Array.from(dependencyMap.entries()).map(([name, path]) => ({ name, path }));
+
+    await deleteFolder(tempFolderPath);
+  }
+
+  async #deleteTypeModuleFromPackageJson(tempFolderPath: string) {
```
vladiulianbogdan, line 273: Is this step still necessary?

src/bundlers/node/nodeJsBundler.ts
```diff
@@ -0,0 +1,336 @@
...
+    const tmpFolder = input.extra.tmpFolder;
+
+    if (mode === "development" && !tmpFolder) {
+      throw new Error("tmpFolder is required in development mode.")
+    }
+
+    const temporaryFolder = mode === "production" ? await createTemporaryFolder() : tmpFolder!;
+    input.extra.dependenciesInfo = [];
+
+    // 1. Run webpack to get dependenciesInfo and the packed file
```
vladiulianbogdan, line 304: Update comment. No webpack anymore.

src/bundlers/node/typeCheckerBundler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    static logChangeDetection() {
+        TypeCheckerBundler.used = false;
+    }
+
+    async bundle(input: BundlerInput): Promise<BundlerOutput> {
+        if (TypeCheckerBundler.used) {
+            return input;
+        }
+        TypeCheckerBundler.used = true;
```
vladiulianbogdan, line 34: Interesting solution. However, while testing, I've noticed that type checker is executed only once for `genezio local`.  If I save the type checker is not executed again. However, I think we want this behaviour. Type checking every time makes `genezio local` very slow. I've noticed the `logChangeDetection` function getting call in local.ts which makes me think that maybe this was not your intention.

src/bundlers/node/typeCheckerBundler.ts
```diff
@@ -0,0 +1,62 @@
...
+                    // Format the diagnostic and write it to stderr
+                    log.error(ts.formatDiagnostic(diagnostic, {
+                        getCanonicalFileName: (fileName) => fileName,
+                        getCurrentDirectory: ts.sys.getCurrentDirectory,
+                        getNewLine: () => ts.sys.newLine,
+                    }));
+                }
+            });
+
+            throw new Error("Typescript syntax error");
```
vladiulianbogdan, line 57: Is it always a syntax error? Maybe it's a typing error. I think we should make this comment more generic.

src/utils/configs.ts
```diff
@@ -1,7 +1,7 @@
 export const tsconfig = {
   compilerOptions: {
     target: "es6",
-    module: "commonjs",
+    // module: "commonjs",
```
vladiulianbogdan, line 4: remove?

src/commands/local.ts
```diff
@@ -301,9 +303,10 @@ function getBundler(
   let bundler: BundlerInterface | undefined;
   switch (classConfiguration.language) {
     case ".ts": {
-      const nodeTsBundler = new NodeTsBundler();
+      const typeCheckerBundler = new TypeCheckerBundler();
+      const nodeJsBundler = new NodeJsBundler();
       const localBundler = new NodeJsLocalBundler();
-      bundler = new BundlerComposer([nodeTsBundler, localBundler]);
+      bundler = new BundlerComposer([typeCheckerBundler, nodeJsBundler, localBundler]);
```
vladiulianbogdan, line 309: I love the fact that we don't have a js and a ts bundler! <3

src/bundlers/node/typeCheckerBundler.ts
costinsin: I have removed the type check on `local` as it was consuming a lot of time.

vladiulianbogdan (approved):

PR #332, opened 2023-06-27
fix: return status code 1 instead of status code 0 for already used p…

Tudor-Ang:
- [ ] 🐛 Bug Fix

## Description
return status code 1 instead of status code 0 for already used ports

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -414,6 +414,7 @@ async function startServerHttp(
       }
 
       reject(error);
+      process.exit(1);
     });
   });
 }


src/commands/local.ts
```diff
@@ -414,6 +414,7 @@ async function startServerHttp(
       }
 
       reject(error);
+      process.exit(1);
```
andreia-oca, line 417: With this line of code, the log `The port 8083 is already in use. Please use a different port by specifying --port <port> to start your local server.` is not shown anymore.

I guess what's happening is that `process.exit` is abruptly killing the NodeJS run loop and the log doesn't have the time to appear. [1]

As suggested on [1], instead of `process.exit(1)` what you can do is set the process exitCode with `process.exitCode = 1;`.

[1] https://mguida.com/blog/exiting-a-node-process/

andreia-oca (changes requested):

PR #333, opened 2023-06-28
Rework temporary folder system so that `genezio local` does not copy …

costinsin:
## Type of change

- [x] 🐛 Bug Fix
- [x] 🧑‍💻 Improvement

## Description

The new system has a root folder named `genezio-${PID}` which contains all temporary folders. At the end of the program we delete this folder for clean up.

The class folders will be named after their class name. This way node_modules is copied only the first time the class is loaded. Other folders will have a random 6 letter character name.

## Checklist

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -33,6 +33,7 @@ import { DartBundler } from "../bundlers/dart/localDartBundler.js";
 import axios, { AxiosResponse } from "axios";
 import { findAvailablePort } from "../utils/findAvailablePort.js";
 import { YamlProjectConfiguration } from "../models/yamlProjectConfiguration.js";
+import hash from 'hash-it';
 
 type ClassProcess = {
   process: ChildProcess;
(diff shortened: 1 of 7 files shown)


src/commands/local.ts
```diff
@@ -261,9 +236,7 @@ async function startProcesses(
 
     debugLogger.log("Start bundling...");
     // TODO: Is it worth the extra complexity of maintaining the folder?
```
vladiulianbogdan, line 238: We can remove this comment.

src/commands/local.ts
```diff
@@ -261,9 +236,7 @@ async function startProcesses(
...
-    return createTemporaryFolder().then(async (tmpFolder) => {
-      temporaryFolders.push(tmpFolder);
-
+    return createTemporaryFolder(classInfo.name).then(async (tmpFolder) => {
```
vladiulianbogdan, line 239: you can replace with `classInfo.name` with a hash over `classInfo.path`.

vladiulianbogdan (approved):

PR #334, opened 2023-06-29
Wait until the ts config file is written successfully

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🐛 Bug Fix

## Description

If a tsconfig file does not exist in the project, the code would try to write the default tsconfig content to the file, but we did not wait for its completion.

Moreover, we had to use target "ES2020" in tsconfig to have support for `import.meta`.

## Checklist

- [x] I have added tests;

### DIFF
--- a/src/utils/configs.ts
+++ b/src/utils/configs.ts
@@ -1,19 +1,27 @@
 export const tsconfig = {
   compilerOptions: {
-    target: "es6",
-    lib: ["es6", "dom"],
-    outDir: "build",
-    rootDir: ".",
-    strict: true,
-    noImplicitAny: true,
-    esModuleInterop: true,
-    resolveJsonModule: true,
-    allowJs: true,
-    types: ["node"]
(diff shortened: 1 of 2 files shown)


src/utils/configs.ts
```diff
@@ -1,19 +1,27 @@
...
-    outDir: "build",
-    rootDir: ".",
-    strict: true,
-    noImplicitAny: true,
-    esModuleInterop: true,
-    resolveJsonModule: true,
-    allowJs: true,
-    types: ["node"]
+      target: "ES2020",
+      module: "commonjs",
```
costinsin, line 4: This option should not be used. In this case, the code generated will be CommonJS. We don't want that as some libraries can't be imported from this type of code.

```suggestion
      module: "NodeNext",
```

costinsin (changes requested):

costinsin (approved):

PR #336, opened 2023-06-29
catch error for existing subdomain

stefandarius:
## Type of change

- [ ] 🍕 New feature
- [ ] 🐛 Bug Fix
- [ ] 🔥 Breaking change
- [x ] 🧑‍💻 Improvement
- [ ] 📝 Documentation Update

## Description

Catch an error to display a friendly message when a user tries to deploy frontend to a taken (by another user) subdomain

## Checklist

- [ ] My code follows the contributor guidelines of this project;
- [ ] I have added tests;
- [ ] New and existing unit tests pass locally with my changes;


radu1122:
why did you change the indentation?

costinsin (approved):
You should update this to use 4 spaces wide tabs. The CLI does not have a rule implied on the width of the tabs. There are files with 2 spaces width and files with 4 spaces width. For code diff, it's better to keep the current indentation, so only the changed code appears in the diff.

costinsin:
In a future update we should impose a tab width rule and we should apply a linter test before merging, but for now we can merge this.

PR #349, opened 2023-07-04
Load variables from `.env` file if it exists in `genezio local`

costinsin:
## Type of change

- [x] 🧑‍💻 Improvement

## Checklist

### DIFF
--- a/src/bundlers/node/lambdaHandler.ts
+++ b/src/bundlers/node/lambdaHandler.ts
@@ -2,6 +2,9 @@
 /* eslint-disable no-useless-escape */
 
 export const lambdaHandler = (className: string): string => `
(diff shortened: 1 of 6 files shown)


src/bundlers/node/lambdaHandler.ts
```diff
@@ -2,6 +2,9 @@
 /* eslint-disable no-useless-escape */
 
 export const lambdaHandler = (className: string): string => `
+/// This is an auto generated code. This code should not be modified since the file can be overwritten
+/// if new genezio commands are executed.
```
andreia-oca, line 6:
```suggestion
/**
* This is an auto generated code. This code should not be modified since the file can be overwritten 
* if new genezio commands are executed.
*/
```

PR #352, opened 2023-07-05
fix: error for classes without methods

Tudor-Ang:
- [ ] 🍕 New feature
- [ ] 🐛 Bug Fix
- [ ] 🔥 Breaking change
- [x] 🧑‍💻 Improvement
- [ ] 📝 Documentation Update

## Description
If we use genezio deploy and we don't have methods in our classes, we display a better error.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -17,7 +17,7 @@ import { ProjectConfiguration } from "../models/projectConfiguration.js";
 import { SdkGeneratorResponse } from "../models/sdkGeneratorResponse.js";
 import { getAuthToken } from "../utils/accounts.js";
 import { getProjectConfiguration } from "../utils/configuration.js";
-import { checkClassesMethods } from '../utils/checkClassesMethods.js'
+import { getNoMethodClasses } from "../utils/getNoMethodClasses.js";
 import {
   fileExists,
   createTemporaryFolder,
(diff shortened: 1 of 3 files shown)


src/commands/deploy.ts
```diff
@@ -181,6 +181,12 @@ export async function deployClasses(configuration: YamlProjectConfiguration, clo
     sdkResponse
   );
 
+  const classesWithEmptyMethods = checkClassesMethods(projectConfiguration.classes);
```
costinsin, line 184: Let's better name the variable and the function name.
Variable could be changed to 'classesWithNoMethods'.

Function name could be renamed to 'getNoMethodClasses()'

src/utils/checkClassesMethods.ts
```diff
@@ -0,0 +1,7 @@
+export const checkClassesMethods = (classes: any) => {
```
costinsin, line 1: Please don't use 'any' type anywhere. Place the expected type.

costinsin, line 1:
```suggestion
export function getNoMethodClasses(classes: any) {
```

I find this more readable.

costinsin (changes requested):

src/utils/getNoMethodClasses.ts
```diff
@@ -0,0 +1,7 @@
+export function getNoMethodClasses(classes: any[]) {
```
costinsin, line 1:
```suggestion
export function getNoMethodClasses(classes: ClassConfiguration[]) {
```

src/utils/getNoMethodClasses.ts
```diff
@@ -0,0 +1,7 @@
...
+  const classesWithNoMethods = classes
+    .filter((obj) => obj.methods.length === 0)
```
costinsin, line 3:
```suggestion
    .filter((obj: ClassConfiguration) => obj.methods.length === 0)
```

src/utils/getNoMethodClasses.ts
```diff
@@ -0,0 +1,7 @@
...
+    .map((obj) => obj.name);
```
costinsin, line 4:
```suggestion
    .map((obj: ClassConfiguration) => obj.name);
```

costinsin (changes requested):
You are still using the `any` type 😂

costinsin (approved):

PR #361, opened 2023-07-10
filter .git file from nonJsFiles and nonDartFiles

Tudor-Ang:
- [x] 🧑‍💻 Improvement

remove .git file from nonJsFiles and nonDartFiles


radu1122 (approved):
Looks good

PR #366, opened 2023-07-20
GNZ-954: Fix for files with wrong perms

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🧑‍💻 Improvement

## Description

We are showing a proper error message now when the user has a file with a permission that does not allow us to copy it.

## Checklist

- [x] I have added tests;


src/bundlers/dart/dartBundler.ts
```diff
@@ -197,7 +199,13 @@ export class DartBundler implements BundlerInterface {
                 }
                 // copy file to tmp folder
                 const fileDestinationPath = path.join(tempFolderPath, filePath.path);
-                return fs.promises.copyFile(filePath.path, fileDestinationPath);
+                return fs.promises.copyFile(filePath.path, fileDestinationPath).catch((error) => {
```
radu1122, line 202: i kind of don't like this solution because if we have multiple files with this problem we will have the same error a lot of times in the console

PR #376, opened 2023-08-03
Install @types modules that are required by TS if they are not already installed.

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🐛 Bug Fix

## Description

If you create a project with `genezio init` and you add a typescript class, you will receive an error because @types/node module is missing. We install it during `genezio local` or `genezio deploy`.

## Checklist

- [x] I have added tests;

### DIFF
--- a/src/bundlers/node/dependencyInstaller.ts
+++ b/src/bundlers/node/dependencyInstaller.ts
@@ -40,7 +40,8 @@ class DependencyInstaller {
                 return;
             }
 
-            const command = "npm install " + (noSave ? "--no-save " : "") + toBeInstalled.join(" ");
+            const baseCommand = this.getInstallPackageCommand(noSave);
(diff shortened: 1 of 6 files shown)


src/bundlers/node/dependencyInstaller.ts
```diff
@@ -54,12 +55,42 @@ class DependencyInstaller {
...
-            const command = "npm install";
+            const command = this.getInstallCommand();
             debugLogger.debug("Running command: " + command);
 
             await exec(command);
         })
     }
+
+    getInstallPackageCommand(noSave?: boolean): string {
+        const dependencyManager = process.env.GENEZIO_DEPENDENCY_MANAGER || "npm";
```
radu1122, line 66: I think this is not really persistent

radu1122 (approved):

PR #394, opened 2023-08-16
add full details to ast summary

stefandarius:
## Type of change

- [x] 🧑‍💻 Improvement

## Description

Make the AST summary more detailed, in order to be able to reconstruct it from the data stored in DB
I added the return type of each function, a `types` vector for each class and an `optional` field for each parameter

## Checklist

- [ ] I have added tests;


src/generateSdk/utils/getAstSummary.ts
```diff
@@ -31,21 +31,30 @@ export function getAstSummary(
         return elem.type === "ClassDefinition";
       }) as ClassDefinition;
 
+      // get the types
```
radu1122, line 34: change everywhere the concept of AstSummary, because now there is full AST

stefandarius:
I also added in this PR the updated `genezio sdk` command (ex `genezio generateSdk`)

This command can be executed anywhere on your system and can do 2 things:
- Generate an SDK for a local project. For this you will need to provide a path to your local project. It can be a full path to your `genezio.yaml` file or just the path to the folder that contains it (it should work either way).
Example: `genezio sdk --source ../../my-project --path ./sdk --port 8083 --language ts`. All options can be omited, default values are `source: ./`, `path: ./sdk`, `port: 8083`, and `language: ts`

- Generate an SDK for a deployed project. For this you will need to know details about your project such as name, stage and region. You can find those by running `genezio ls` and looking for your project.
Example: `genezio sdk my-project --stage prod --region us-east-1`. The project name is mandatory, the options have default values (`region: us-east-1`, `stage: prod`). You can also use `--path` and `--languange` here. If you provide an existing project name but the stage and region don't match with any deployed project, you will recieve an error.

PR #425, opened 2023-09-14
catch errors in setTimeout and send async errors to sentry dashboard

stefandarius:
## Type of change

- [x] 🧑‍💻 Improvement

## Description

Updated handler to catch errors sent inside a setTimeout callback. Also added Sentry logic in more places, to be able to catch and send error sent in async functions.

## Checklist

### DIFF
--- a/src/bundlers/node/lambdaHandler.ts
+++ b/src/bundlers/node/lambdaHandler.ts
@@ -172,8 +172,20 @@ if (!genezioClass) {
 
             const requestId = body.id;
             const errorPromise = new Promise((resolve) => {
-                process.on("uncaughtException", function (err) {
+                process.removeAllListeners("uncaughtException")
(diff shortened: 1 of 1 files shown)


src/bundlers/node/lambdaHandler.ts
```diff
@@ -172,8 +172,20 @@ if (!genezioClass) {
 
             const requestId = body.id;
             const errorPromise = new Promise((resolve) => {
-                process.on("uncaughtException", function (err) {
+                process.removeAllListeners("uncaughtException")
+                process.on("uncaughtException", async function (err) {
                     console.error(err);
+                    try {
```
vladiulianbogdan, line 178: Should we create a function with this?

vladiulianbogdan (approved):

PR #502, opened 2023-10-30
fix genezio super cmd

radu1122:
# Pull Request Template

## Type of change

- [ ] 🍕 New feature
- [x] 🐛 Bug Fix
- [ ] 🔥 Breaking change
- [x] 🧑‍💻 Improvement
- [ ] 📝 Documentation Update

## Checklist

- [ ] My code follows the contributor guidelines of this project;
- [ ] I have added tests;
- [ ] New and existing unit tests pass locally with my changes;

### DIFF
--- a/src/genezio.ts
+++ b/src/genezio.ts
@@ -7,7 +7,7 @@ import {
 } from "./constants.js";
 import log from "loglevel";
 import prefix from 'loglevel-plugin-prefix';
-
+import latestVersion from "latest-version";
 // commands imports
 import { accountCommand } from "./commands/account.js";
 import { addClassCommand } from "./commands/addClass.js";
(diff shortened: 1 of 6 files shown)


src/genezio.ts
```diff
@@ -47,24 +46,37 @@ if (ENABLE_DEBUG_LOGS_BY_DEFAULT) {
...
+
+  await genezioCommand().catch((error: Error) => {
+    log.error(error.message);
+    GenezioTelemetry.sendEvent({eventType: TelemetryEventTypes.GENEZIO_COMMAND_ERROR, errorTrace: error.message});
+    exit(1);
+  });
+}
+
+// make genezio --version
+program.version(await latestVersion("genezio"), "-v, --version", "Output the current version of genezio.");
```
andreia-oca, line 64: I think this is not the version of genezio installed on the user's machine. This is the latest from `npm`.

You can use `version` from our `utils.ts`.

src/utils/configs.ts
```diff
@@ -43,3 +43,45 @@ export const regions = [
...
+"Canada (Central)",
+"Europe (Frankfurt)",
+"Europe (Ireland)",
+"Europe (London)",
+"Europe (Paris)",
+"Europe (Stockholm)",
+"South America (São Paulo)"
+];
+
+export const regionsNames = [
```
andreia-oca, line 67: This seems unused. Can we removed it?

I feel it will create confusion in the future if we leave it here.

andreia-oca (approved):
I have two inline comments. Please fix them, then imo this can be merged.

PR #508, opened 2023-11-01
Only one genezio.yaml configuration file for workspaces.

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🍕 New feature

## Description

Instead of having a `genezio.yaml` in the root of the workspace and two `genezio.yaml`s in the server and client, now we are having only one `genezio.yaml` in the root that contains information about both server and client.

I removed some of the code that was parsing the root `genezio.yaml` in a different way than the other two yamls. Also, I've modified the writeToFile method that was having different behaviours based on what type of yaml it is.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -597,7 +597,8 @@ export async function deployFrontend(
       );
     }
     // check if the build folder exists
-    const frontendPath = path.join(cwd, configuration.frontend?.path);
+    // const frontendPath = path.join(cwd, configuration.frontend?.path);
+    const frontendPath = configuration.frontend?.path
     if (!(await fileExists(frontendPath))) {
       throw new Error(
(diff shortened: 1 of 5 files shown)


src/commands/deploy.ts
```diff
@@ -597,7 +597,8 @@ export async function deployFrontend(
       );
     }
     // check if the build folder exists
-    const frontendPath = path.join(cwd, configuration.frontend?.path);
+    // const frontendPath = path.join(cwd, configuration.frontend?.path);
```
costinsin, line 600: Remove this

costinsin:
Can you please further explain what was the behavior before this change and what is the current behavior now? It would make me better understand the change.

andreia-oca (approved):

vladiulianbogdan:
> Can you please further explain what was the behavior before this change and what is the current behavior now? It would make me better understand the change.

I've updated the description.

PR #509, opened 2023-11-01
Improve `login` and `logout` commands

costinsin:
This change handles 2 edge cases:
- When `login`, if there is already a token in the `.npmrc` file, the old token, is just replaced by the new one.
- When `logout`, instead of deleting the entire `.npmrc` file, delete only the genzio registry in case the user has other custom registries that he doesn't want to remove.

# Pull Request Template

## Type of change

- [x] 🐛 Bug Fix

## Checklist


src/utils/accounts.ts
```diff
@@ -24,17 +24,21 @@ function getNpmConfigFileContent(token: string) {
...
     const npmConfigFile = ".npmrc"
-    const npmConfigContent = getNpmConfigFileContent(token);
+    let npmConfigContent = getNpmConfigFileContent(token);
 
     if (fs.existsSync(path.join(homeDirectory, npmConfigFile))) {
         const npmConfigFileContent = await readUTF8File(path.join(homeDirectory, npmConfigFile));
         if (npmConfigFileContent.includes(getNpmConfigFileContent(""))) {
-            return;
+            npmConfigContent = npmConfigFileContent.replace(
+                new RegExp(`(@genezio-sdk:registry=https://${GENEZIO_REGISTRY}/npm\n//${GENEZIO_REGISTRY}/:_authToken=)(.*)(\n?)`),
```
radu1122, line 33: I would remove `@genezio-sdk:registry=https://${GENEZIO_REGISTRY}/npm\n`

src/utils/accounts.ts
```diff
@@ -60,15 +64,16 @@ export async function removeAuthToken(): Promise<[void, void]> {
...
-    const npmrcPromise: Promise<void> = new Promise((resolve, reject) => {
-        fs.unlink(npmConfigFilePath, (error) => {
-            if (error) {
-                reject(error)
-            }
+    let updateNpmConfigPromise = Promise.resolve();
+    if (fs.existsSync(npmConfigFilePath)) {
+        const configContent = await readUTF8File(npmConfigFilePath);
+        const configContentWithoutRegistry = configContent.replace(
+            RegExp(`@genezio-sdk:registry=https://${GENEZIO_REGISTRY}/npm\n//${GENEZIO_REGISTRY}/:_authToken=.*\n?`),
```
radu1122, line 71: same here

src/utils/accounts.ts
vladiulianbogdan: Why? How would that work?

vladiulianbogdan (approved):

src/utils/accounts.ts
costinsin: Done

src/utils/accounts.ts
costinsin: @vladiulianbogdan The lines do not necessarily come one after another. I have made a commit that treats this case. Now, the lines are deleted independently and the update is only done only on the `//${GENEZIO_REGISTRY}/:_authToken=.*` line.

PR #513, opened 2023-11-02
Worker compiler threads

stefandarius:
## Type of change

- [x] 🧑‍💻 Improvement

## Description

SDK compilation was taking too long to perform two times (for CJS and ESM). I used worker threads to parallelize this operation and reduce time.
When compiling for a local SDK, we also perform the `npm link` command at the same time in the main thread, but when we compile for deploy, we need to wait for the compile to finish in order to run `npm publish`

## Checklist


src/commands/local.ts
```diff
@@ -321,50 +328,7 @@ export async function startLocalEnvironment(options: GenezioLocalOptions) {
...
-  const modulePath = path.resolve(sdkPath, "..", "genezio-sdk");
-  await writeToFile(
-    modulePath,
-    "package.json",
-    getNodeModulePackageJsonLocal(projectName, region),
-  );
-  await asyncExec(packageManager + " link", { cwd: modulePath });
-}
-
+// async function compileSdk
```
vladiulianbogdan, line 331: remove this comment

vladiulianbogdan (approved):

PR #567, opened 2023-11-24
Remove npm link

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🧑‍💻 Improvement

## Description

When the user is testing locally, genezio was using `npm link` to install the latest SDK from the local disk. However, we've seen multiple issues with `npm link`:

- the package is removed if the user is running `npm install` or `npm install <package>`
- the user does not understand or forgets to run the `npm link` command
- if `npm` is installed with `nvm` multiple versions of `node` can cause problems

Instead of relying on `link`, we have decided to write directly to `node_modules/@genezio-sdk/...`. We are watching the folder, to see if the package is removed and we put it back if that happens. If `genezio deploy` is executed, we stop all the `genezio local` processes to make sure that the user doesn't push an SDK with localhost URL.

## Checklist

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -27,6 +27,7 @@ import {
     getBundleFolderSizeLimit,
     readEnvironmentVariablesFile,
     createLocalTempFolder,
+    interruptLocalProcesses,
 } from "../utils/file.js";
 import { printAdaptiveLog, debugLogger } from "../utils/logging.js";
 import { runNewProcess } from "../utils/process.js";
@@ -45,12 +46,15 @@ import { GenezioTelemetry, TelemetryEventTypes } from "../telemetry/telemetry.js
 import { TsRequiredDepsBundler } from "../bundlers/node/typescriptRequiredDepsBundler.js";
 import { setEnvironmentVariables } from "../requests/setEnvironmentVariables.js";
 import colors from "colors";
+import os from "os";
+import fs from "fs/promises";
 import { getEnvironmentVariables } from "../requests/getEnvironmentVariables.js";
 import { getNodeModulePackageJson } from "../generateSdk/templates/packageJson.js";
 import { getProjectEnvFromProject } from "../requests/getProjectInfo.js";
 import { compileSdk } from "../generateSdk/utils/compileSdk.js";
 
 export async function deployCommand(options: GenezioDeployOptions) {
+    await interruptLocalProcesses();
     let configuration;
 
     try {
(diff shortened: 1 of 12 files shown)


src/commands/deploy.ts
```diff
@@ -45,12 +46,15 @@ import { GenezioTelemetry, TelemetryEventTypes } from "../telemetry/telemetry.js
 import { TsRequiredDepsBundler } from "../bundlers/node/typescriptRequiredDepsBundler.js";
 import { setEnvironmentVariables } from "../requests/setEnvironmentVariables.js";
 import colors from "colors";
+import os from "os";
+import fs from "fs/promises";
```
costinsin, line 50: Are those used anywhere in this file?

src/utils/file.ts
```diff
@@ -404,3 +405,12 @@ export async function readEnvironmentVariablesFile(
     }
     return envVars;
 }
+
+// Inform local processes to interrupt when a deployment has started
+export async function interruptLocalProcesses() {
+    const interruptLocalPath = path.join(os.homedir(), ".geneziointerrupt");
+    const interruptLocal = await fsPromise.open(interruptLocalPath, "w");
+    await interruptLocal.close();
```
costinsin, line 413: 1. What happens if there is no such path?
2. Why open and close if you modify the access time below?

src/utils/linkDatabase.ts
```diff
@@ -0,0 +1,62 @@
+import { promises as fs } from "fs";
+import os from "os";
+import path from "path";
+
+async function getLinkContent(): Promise<Map<string, string[]>> {
+    const filePath = path.join(os.homedir(), ".geneziolinks");
```
costinsin, line 6: I'm worried that we are creating too many `.genezio(.*)` files. I propose to have a `.genezio` directory that stores all this helper files to avoid polluting the user home directory.

src/commands/local.ts
```diff
@@ -343,6 +363,138 @@ export async function startLocalEnvironment(options: GenezioLocalOptions) {
...
+    }
+}
+
+async function watchNodeModules(
+    yamlProjectConfiguration: YamlProjectConfiguration,
+    sdkPath: string,
+): Promise<NodeJS.Timeout> {
+    // We are watching for the following files:
+    // - node_modules/@genezio-sdk/<projectName>_<region>/package.json: this file is used to determine if the SDK was changed (by a npm install or npm update)
+    // - node_modules/.package-lock.json: this file is used by npm to determine if it should update the packages or not. We are removing this file while "genezio local"
```
costinsin, line 376: Keep in mind that this is incompatible with yarn and pnpm. You can leave a TODO comment if you want to approach this later. 

Pnpm is using a `.pnpm` file and yarn is using a `.yarn.state` file.

costinsin (approved):

PR #576, opened 2023-11-29
Local docker db

Virgil993:
# Pull Request Template

## Type of change

- [x] 🍕 New feature
- [ ] 🧑‍💻 Improvement

## Description

This PR adds the local db feature, it allows the user to change the yaml so that they can set a database like postgres or redis and when they run `genezio local`, a docker container with a local database will be opened and the user will receive the connection string so that he may connect. The docker container will be closed when the genezio local procces is closed.
To use this feature, update your genezio yaml with a new property database like this:
//genezio.yaml
...
```
database: 
   type: postgres
```
    
Modify the type to postgres or redis according to what you need

List any dependencies that are required for this change.

## Checklist

- [ ] I have added tests;


andreia-oca (approved):
Looks good, great job doing this feature - I am excited to see it in action!

I have a few user experience improvements in comments.

src/commands/local.ts
```diff
@@ -610,6 +629,174 @@ function getBundler(classConfiguration: ClassConfiguration): BundlerInterface |
...
+
+export async function stopDockerDatabase() {
+    const yamlProjectConfiguration = await getProjectConfiguration();
+    if (yamlProjectConfiguration.database?.type) {
+        try {
+            await asyncExec("docker --version");
+        } catch (error: any) {
+            return undefined;
+        }
+        switch (yamlProjectConfiguration.database?.type) {
```
radu1122, line 777: i feel like this is duplicated and not needed

src/index.ts
```diff
@@ -30,10 +33,17 @@ process.on("SIGINT", async () => {
         errorTrace: "",
         commandOptions: "",
     });
+    if (currentCommand == "local") {
```
radu1122, line 36: i think would be nicer to have currentCommand at process.env.currentCommand

src/commands/local.ts
```diff
@@ -93,6 +97,18 @@ export async function prepareLocalEnvironment(
                 throw new Error(GENEZIO_NO_CLASSES_FOUND);
             }
 
+            // Start docker container
+            if (yamlProjectConfiguration.database) {
+                printAdaptiveLog(
+                    `Starting local docker ${yamlProjectConfiguration.database?.type} database`,
+                    "start",
+                );
+                await startDockerDatabase(
```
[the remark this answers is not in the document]
costinsin: I would like to have the option to keep the same database between reloads. And I think this should be default.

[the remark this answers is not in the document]
Virgil993: this starts the container only if it isn't running, which is an edge case if the user manually closed his container while a local is running.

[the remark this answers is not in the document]
costinsin: Oh okey! This seems right to me!

src/commands/local.ts
```diff
@@ -695,7 +700,9 @@ async function startDockerDatabase(database: YamlDatabaseConfiguration, projectN
                 "postgres",
                 projectName,
                 `docker run -p 5432:5432 --name genezio-postgres-${projectName} -e POSTGRES_PASSWORD=mysecretpassword -d postgres `,
-                `Local Postgres URL = postgres://postgres:mysecretpassword@localhost:5432`,
+            );
+            log.info(
+                `Your local postgres database is up and running.\nYou can connect to your postgres instance using the following URL: at postgres://postgres:mysecretpassword@localhost:5432`,
```
costinsin, line 705:
```suggestion
                `Your local postgres database is up and running.\nYou can connect to your postgres instance using the following URL: postgres://postgres:mysecretpassword@localhost:5432`,
```

src/commands/local.ts
```diff
@@ -610,6 +629,174 @@ function getBundler(classConfiguration: ClassConfiguration): BundlerInterface |
     return bundler;
 }
 
+async function runDockerContainer(databaseType: string, projectName: string, runCommand: string) {
```
vladiulianbogdan, line 632: can you move all the functions related to docker management in their own file?

vladiulianbogdan, line 632: the name of the function suggests that it starts a general purpose docker, but what it actually does is that it starts a docker database. name it accordingly. one suggestion would be `startLocalDatabaseContainer` but you can also think of something else.

src/commands/local.ts
vladiulianbogdan: I think what @radu1122 is saying is that all the cases are doing the same thing. What was the intention?

vladiulianbogdan:
> I would use https://www.npmjs.com/package/dockerode instead of running commands on the user machine.
> 
> What if I `alias docker=dajsdhasdjk`? You are assuming that I have the docker CLI installed. The docker daemon can run without the CLI to be installed.

Totally agree.

costinsin (changes requested):
My opinion is that we should not further introduce new `any`s in the code, even if they are used in `catch` statements. Let's try checking if the error received is an instance of the Error class at runtime.

I have also sugested removing some code that is no longer used.

src/utils/localDockerDatabase.ts
```diff
@@ -0,0 +1,201 @@
...
+                currentContainer = dockerHandler.getContainer(container.Id);
+            }
+        });
+
+        // output = await asyncExec(`docker ps -a --format "{{.Names}}" `);
+    } catch (error: any) {
+        printAdaptiveLog(`Failed starting local docker ${databaseType} database`, "");
+        log.error(`An error has occured: ${error.toString()}`);
+        log.error(
+            "This is most likely a docker error, check if your docker daemon is running.\nIf it is not running you can learn how to start it by going to the docker documentation at https://docs.docker.com/ ",
```
costinsin, line 27:
```suggestion
    } catch (error) {
        printAdaptiveLog(`Failed starting local docker ${databaseType} database`, "");
        if (error instanceof Error) {
            log.error(`An error has occured: ${error.toString()}`);
        }
        log.error(
            "This is most likely a docker error, check if your docker daemon is running.\nIf it is not running you can learn how to start it by going to the docker documentation at https://docs.docker.com/ ",
```

src/utils/localDockerDatabase.ts
```diff
@@ -0,0 +1,201 @@
...
```
costinsin, line 165:
```suggestion
    } catch (error) {
        printAdaptiveLog(`Failed starting local docker ${databaseType} database`, "");
        if (error instanceof Error) {
            log.error(`An error has occured: ${error.toString()}`);
        }
        log.error(
            "This is most likely a docker error, check if your docker daemon is running.\nIf it is not running you can learn how to start it by going to the docker documentation at https://docs.docker.com/ ",
```

costinsin (changes requested):
Two more `any`s :P

src/utils/localDockerDatabase.ts
```diff
@@ -0,0 +1,224 @@
...
+                    });
+                    if (!isImagePresent) {
+                        const pullStream = await dockerHandler.pull("postgres");
+                        await new Promise((res) =>
+                            dockerHandler.modem.followProgress(pullStream, res),
+                        );
+                    }
+                    const newContainer = await dockerHandler.createContainer({
+                        Image: "postgres",
+                        Env: [`POSTGRES_PASSWORD=mysecretpassword`],
```
costinsin, line 64: I was thinking that the container should inherit the local process ENV. I am not sure about this, you don't need to change anything for now, just keep in mind that maybe we should think about this and change it in the future.

costinsin (approved):
The code LGTM (looks good to me)! Thank you for doing a good job!

PS: Please update this before merging https://github.com/Genez-io/genezio/pull/576#discussion_r1419088507

andreia-oca (approved):
Looks great! Ready to be merged! 🚀

PR #577, opened 2023-11-29
Fix broken links in .npmrc

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🐛 Bug Fix

## Description

The NPM registry url that was written during `genezio login` was incomplete.

### DIFF
--- a/src/packageManagers/npm.ts
+++ b/src/packageManagers/npm.ts
@@ -35,19 +35,21 @@ export default class NpmPackageManager implements PackageManager {
 
         // Add the authentication token for the registry hostname
         const registryUrl = new URL(url);
-        await asyncExec(`npm config set //${registryUrl.hostname}/:_authToken=${authToken}`);
+        const path = registryUrl.pathname.split("/").slice(0, -1).join("/");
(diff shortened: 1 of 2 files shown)


src/packageManagers/npm.ts
```diff
@@ -35,19 +35,21 @@ export default class NpmPackageManager implements PackageManager {
 
         // Add the authentication token for the registry hostname
         const registryUrl = new URL(url);
-        await asyncExec(`npm config set //${registryUrl.hostname}/:_authToken=${authToken}`);
+        const path = registryUrl.pathname.split("/").slice(0, -1).join("/");
+        await asyncExec(`npm config set //${registryUrl.hostname}${path}/:_authToken=${authToken}`);
```
costinsin, line 39: Don't you need a slash between the hostname and the path?

Like: ${hostname}/${path}

this comment applies to all package managers.

costinsin (changes requested):

src/packageManagers/npm.ts
vladiulianbogdan: No. It already has one it in the path.

costinsin: ok

costinsin (approved):

PR #578, opened 2023-11-30
Add support for genezio middlewares.

vladiulianbogdan:
# Pull Request Template

## Type of change

- [x] 🍕 New feature

## Description

Add support for genezio middlewares. Now we can do something like

```
function GenezioInputInspector(callback: (...args: any[]) => void) {
    return function (value: any, context: any) {
        return async function (...args: any[]) {
            callback(...args)
            const result =  await value(...args);
            return result
        };
    };
}
```

```
@GenezioDeploy()
export class HelloWorld {

  constructor() {
    console.log("Constructor called!")
  }
  
  @GenezioAuth()
  helloAuth(context: GenezioContext) {
      console.log("context is ", context)
  }

  /**
  * Method that returns a "Hello world" message.
  */
  @GenezioInputInspector((x) => { console.log("New request received with parameters", x) })
  helloWorld(test: string) {
    console.log("Hello world request received!", test)

    return "Hello world!";
  }

  /**
   * Method that returns a personalized "Hello world" message.
   */
  @GenezioInputInspector((name, from, value) => {
    z.string().parse(name);
    z.string().parse(from);
    z.nativeEnum(Season).parse(value)
  })
  hello(name: string, from: string, value: Season): string {
    console.log(`Hello world request received with name ${name} from ${from} value ${value}!`)

    const message = `Hello, ${name}, from ${from} during this ${value}`;
    console.log(message)

    return message
  }
}
```


costinsin (changes requested):

costinsin:
Just to be clear, this change does not add support for `GenezioInputInspector` or `GenezioAuth`, right? As I understand, it just stops striping decorators and updates to the latest proposal.

vladiulianbogdan:
> Just to be clear, this change does not add support for `GenezioInputInspector` or `GenezioAuth`, right? As I understand, it just stops striping decorators and updates to the latest proposal.

Yes.

src/utils/configuration.ts
```diff
@@ -48,9 +48,10 @@ function parseArguments(args: any): { [key: string]: string } {
                     return [...acc, { key: curr.key.name, value: curr.value.value }];
                 }, []);
             } else {
-                throw new Error("Unsupported argument for decorator");
+                return undefined;
             }
         })
+        .filter((a: any) => a !== undefined)
```
costinsin, line 54: Please use unknown instead of any.
```suggestion
        .filter((a: unknown) => a !== undefined)
```

costinsin (approved):

PR #619, opened 2023-12-11
Bugfix/stop checking more than layer one folders when running genezio super cmd

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
-   [x] 🧑‍💻 Improvement

## Description

There was a performance issue when running genezio super command. To check if the directory was empty when running genezio super command we would check recursively. This would mean that if i would run genezio super command in a directory with a lot of other directories and big projects then the command would take a lot of time to go through all of them. Now, the genezio super command only checks the first layer.

## Checklist


andreia-oca (approved):
Looks good.

PR #637, opened 2023-12-13
Add a workflow to automatically push to npm on a new tag

andreia-oca:
# Pull Request Template

## Type of change

-   [x] 🧑‍💻 Improvement

## Description

This PR automatically pushes to  the npm registry.

There are 2 ways of triggering this workflow:
- when a version tag is pushed to the repo
- manually

The developer still needs to manually increment the package version in an explicit PR/commit push to `main`.

This will ease the process of releasing the cli tool and make it less error prone.

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,24 +1,25 @@
+# This workflow is triggered when a tag v* is pushed to the repository
+
+# To release to npm, make sure you incremented the version in package.json
+# and then run `git tag vx.x.x` and `git push origin vx.x.x`
 name: npm-publish
 on:
     push:
-        branches:
-            - main
+        tags:
+            - "v*"
+    workflow_dispatch:
+
 jobs:
-    npm-publish:
-        name: npm-publish
+    publish:
         runs-on: ubuntu-latest
         steps:
-            - name: Checkout repository
-              uses: actions/checkout@v2
-            - name: Publish if version has been updated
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,24 +1,25 @@
...
+
 jobs:
-    npm-publish:
-        name: npm-publish
+    publish:
         runs-on: ubuntu-latest
         steps:
-            - name: Checkout repository
-              uses: actions/checkout@v2
-            - name: Publish if version has been updated
```
costinsin, line 13: Don't we want to keep this?

costinsin (approved):

.github/workflows/release.yaml
costinsin: I mean, a check that npm version was incremented

andreia-oca: Nope, I am taking a more lightweight approach to publishing to npm - as seen in GitHub documentation [1].

Afaik, this action I deleted does some magic to recognise a version change in the `package.json` and triggers an npm publish - which is nice, but didn't previously worked.

My proposal is to trigger the publishing when pushing a tag or manually by clicking the run button in github, so no more magic.

[1] https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

PR #650, opened 2023-12-15
Support for using genezio CLI for typesafe calls

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

We aim to integrate genezio into existing applications by adding genezio classes to an express.js path. This allows us to utilize the genezio SDK for type-safe method calls.

This PR includes:

- An update to the genezio sdk command, enabling SDK generation for annotated classes without requiring a genezio project.
- Modifications to the TypeScript and JavaScript SDKs to prevent Vite warnings related to dynamic imports.


src/models/commandOptions.ts
```diff
@@ -27,12 +29,19 @@ export interface GenezioDeleteOptions extends BaseOptions {
     force: boolean;
 }
 
+export enum SourceType {
+    local = "local",
+    remote = "remote",
```
costinsin, line 34:
```suggestion
    LOCAL = "local",
    REMOTE = "remote",
```

src/genezio.ts
```diff
@@ -229,12 +228,18 @@ program
     )
     .option("--language <language>", "Language of the SDK.", "ts")
     .option(
-        "-s, --source <source>",
+        "--source <source>",
+        "Where the SDK should be generated from. Possible values: local, remote.",
+        "./",
```
costinsin, line 233:
```suggestion
    .addOption(
        new Option("--source <source>", "Where the SDK should be generated from.")
            .choices(["local", "remote"])
            .default("local"),
    )
```

This way, you can specifiy the possible choices and they will be handled by commander. Also you should set `local` or `remote` (not sure which, you decide) by default, not `"./"`.

src/commands/generateSdk.ts
```diff
@@ -23,9 +23,61 @@ import { mapDbAstToSdkGeneratorAst } from "../generateSdk/utils/mapDbAstToFullAs
...
+    switch (options.source) {
+        case SourceType.local:
+            await generateLocalSdkCommand(options);
+            break;
+        case SourceType.remote:
+            await generateRemoteSdkCommand(projectName, options);
+            break;
+        default:
+            log.error("Invalid source type.");
+            exit(1);
```
costinsin, line 39: No longer needed if you let `Commander` handle it.

src/commands/generateSdk.ts
```diff
@@ -23,9 +23,61 @@ import { mapDbAstToSdkGeneratorAst } from "../generateSdk/utils/mapDbAstToFullAs
...
+    }
+}
+
+export async function generateLocalSdkCommand(options: GenezioSdkOptions) {
+    let configuration: YamlProjectConfiguration = await YamlProjectConfiguration.create({
+        name: "test",
```
costinsin, line 45: Is this `test` name intended?

src/genezio.ts
```diff
@@ -229,12 +228,18 @@ program
...
+    )
+    .option(
+        "-c, --config <config>",
         "Path to the genezio.yaml file on your disk. Used for loading project details from a genezio.yaml file, instead of command argumments like --name",
         "./",
     )
     .option("-p, --path <path>", "Path to the directory where the SDK will be generated.", "./sdk")
     .option("--stage <stage>", "Stage of the project.", "prod")
+    .option("--url <url>", "The url of the server.")
```
costinsin, line 242:
```suggestion
    .requiredOption("--url <url>", "The url of the server.")
```

This option has no default, but it was defined as it can't be undefined in the `GenezioSdkOptions` type. Please make it required.

costinsin (changes requested):

andreia-oca (approved):
I agree with Costin's comments.

Other than that, looks good to me.

A documentation update should also be made for the cli page:
https://docs.genez.io/genezio-dev-documentation/cli-tool/cli-commands/generatesdk

src/commands/generateSdk.ts
vladiulianbogdan: Yea, it's kind of a hack because I am forced to create a `YamlProjectConfiguration` object and I have to provide a name.

costinsin (approved):

PR #683, opened 2024-01-03
add index file to old ts and js sdk generation

stefandarius:
# Pull Request Template

## Type of change

-   [x] 🧑‍💻 Improvement

## Description

We added the index file (a file that exports all the types and classes of an sdk) in order to have an entry point for our genezio sdk node module, so initially we only added it for the new way of generating the sdk.
Now it is clear that is much cleaner to import everything from one place, so from now on we add this file in the old way of generating the sdk as well. This should not break any existing genezio projects as the old way of importing types and classes still works.

## Checklist

### DIFF
--- a/src/models/genezioModels.ts
+++ b/src/models/genezioModels.ts
@@ -356,8 +356,5 @@ export enum SdkVersion {
  * A class implementing this interface will create the sdk for a given language.
  */
 export interface SdkGeneratorInterface {
-    generateSdk: (
-        sdkGeneratorInput: SdkGeneratorInput,
-        sdkVersion: SdkVersion,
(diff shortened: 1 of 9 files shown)


src/models/genezioModels.ts
```diff
@@ -356,8 +356,5 @@ export enum SdkVersion {
  * A class implementing this interface will create the sdk for a given language.
  */
 export interface SdkGeneratorInterface {
-    generateSdk: (
-        sdkGeneratorInput: SdkGeneratorInput,
-        sdkVersion: SdkVersion,
```
radu1122, line 361: should also delete the model SdkVersion

vladiulianbogdan (approved):

PR #694, opened 2024-01-08
Set up a/b testing for capybara runtime

MagoDelBlocco:
🔥 Breaking change

## Description

Add a random choice for the default cloud provider, in order to a/b test the genezio capybara cloud provider. It will have a 50/50 chance of selecting capybara or genezio as cloud providers, if the environment is "dev". Additionally, if this choice is made, the genezio.yaml file will be overwritten, in order to preserve this choice.

### DIFF
--- a/src/models/yamlProjectConfiguration.ts
+++ b/src/models/yamlProjectConfiguration.ts
@@ -7,6 +7,7 @@ import { CloudProviderIdentifier } from "./cloudProviderIdentifier.js";
 import { DEFAULT_NODE_RUNTIME, NodeOptions } from "./nodeRuntime.js";
 import zod from "zod";
 import log from "loglevel";
+import { ENVIRONMENT } from "../constants.js";
 
 export enum TriggerType {
     jsonrpc = "jsonrpc",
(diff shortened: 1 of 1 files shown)


src/models/yamlProjectConfiguration.ts
```diff
@@ -462,6 +472,12 @@ export class YamlProjectConfiguration {
             workspace,
             configurationFile.packageManager,
         );
+        if (ENVIRONMENT === "dev") {
+            // Used for a/b testing of genezio runtime.
+            // TODO: remove after a/b testing is complete.
+            yamlProjectConfiguration.writeToFile();
```
valighita, line 478: This should be done only if we change the provider.

MagoDelBlocco: I doubt `zod` gives a function to determine if the default option has been selected or not.

src/models/yamlProjectConfiguration.ts
```diff
@@ -312,6 +313,15 @@ export class YamlProjectConfiguration {
...
 
+        // Used for a/b testing of genezio runtime.
+        // TODO: Remove after a/b testing is complete, and use a determined cloud provider.
+        const cloudProviderDefault =
+            ENVIRONMENT === "dev"
+                ? Math.random() < 0.5
+                    ? CloudProviderIdentifier.GENEZIO
+                    : CloudProviderIdentifier.CAPYBARA_LINUX
+                : CloudProviderIdentifier.GENEZIO;
+
```
andreia-oca, line 324: This approach has multiple disadvantages:
- Most of our configuration files are generated automatically and they already are setting the cloud provider to `genezio`, so this code will take effect rarely.
- This code will be executed for every genezio command, so for example, I can trigger a cloud provider change by running `genezio list` or `genezio local` which is not desirable.
- We don't want to change cloud provider for a project that is already deployed because currently we don't support changing providers on the fly.

I played around and come up with an approach on the deployment flow that solves the concerns above:
https://github.com/Genez-io/genezio/pull/818

MagoDelBlocco:
Superseded by https://github.com/Genez-io/genezio/pull/818

PR #699, opened 2024-01-10
Make genezio deploy, account and list automatically login if token is incorrect

costinsin:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description
This makes the user flow easier because we login automatically for him if his token is missing or incorrect.

## Checklist

### DIFF
--- a/src/commands/superGenezio/superGenezio.ts
+++ b/src/commands/superGenezio/superGenezio.ts
@@ -1,8 +1,5 @@
 import path from "path";
-import { getAuthToken } from "../../utils/accounts.js";
 import { fileExists } from "../../utils/file.js";
-import { loginCommand } from "../login.js";
-import { debugLogger } from "../../utils/logging.js";
 import { deployCommand } from "../deploy.js";
 import colors from "colors";
 import inquirer from "inquirer";
@@ -47,11 +44,6 @@ export async function genezioCommand() {
                 const options: GenezioDeployOptions = {
                     installDeps: true,
                 };
-
-                if (!(await getAuthToken())) {
-                    debugLogger.debug("No auth token found. Starting automatic authentication...");
-                    await loginCommand("", false);
-                }
 
                 return await deployCommand(options).catch(async (error) => {
                     await GenezioTelemetry.sendEvent({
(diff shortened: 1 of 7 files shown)


costinsin:
Current dependencies on/for this PR:
* `dev`
  * **PR #699** <a href="https://app.graphite.dev/github/pr/Genez-io/genezio/699?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Genez-io/genezio/699?utm_source=stack-comment).

src/commands/superGenezio/superGenezio.ts
```diff
@@ -48,11 +45,6 @@ export async function genezioCommand() {
                     installDeps: true,
                 };
 
-                if (!(await getAuthToken())) {
```
radu1122, line 51: why did you remove this?

src/utils/accounts.ts
```diff
@@ -18,6 +19,15 @@ export async function getAuthToken(): Promise<string | undefined> {
     }
 }
 
+export async function isLoggedIn(): Promise<boolean> {
+    try {
+        await getUser();
```
radu1122, line 24: is this working? `getUser()` will throw err if there is no user connected?

src/commands/superGenezio/superGenezio.ts
costinsin: Because the auto login is done now in the `deployCommand` function. We no longer need to auto login here.

radu1122: let s also keep it here

src/utils/accounts.ts
costinsin: Yes, it will throw `Error(GENEZIO_NOT_AUTH_ERROR_MSG)` if no token was found

radu1122: yes, but there is also the case when there is a token. but it is invalid

costinsin: I tested it, stop making guesses please

radu1122 (approved):

PR #701, opened 2024-01-11
Bugfix sourcemaps not generating properly due to babel transform

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
-   [x] 🧑‍💻 Improvement

## Description

There was a bug that in the logs from the dashboard, the error stack trace didn't show the original error location. This was due to the fact that the sourcemap, which should have been in the bundled file, was overwritten by the babel transformation. This PR fixes this by setting the sourcemap style to be inline in the babel transformation. Also, due to the fact that the eslint was configured to have scripts in bash, the eslint often stopped Windows users from commiting anything to the Repo. I changed that by making the script in native JS so that they would be OS agnostic.

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -16,8 +16,6 @@ module.exports = {
     ],
     env: {
         node: true,
-        commonjs: true,
-        browser: true,
         es6: true,
     },
 };
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -14,4 +14,10 @@ module.exports = {
         "solveEnvVariables.js",
         ".eslintrc.cjs",
     ],
+    env: {
+        node: true,
+        commonjs: true,
+        browser: true,
+        es6: true,
+    },
 };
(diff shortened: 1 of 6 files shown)


radu1122 (approved):

.eslintrc.cjs
```diff
@@ -14,4 +14,10 @@ module.exports = {
         "solveEnvVariables.js",
         ".eslintrc.cjs",
     ],
+    env: {
+        node: true,
+        commonjs: true,
+        browser: true,
```
costinsin, line 20: Why are we enabling these 2?

package.json
```diff
@@ -12,8 +12,8 @@
         "build": "npx tsc && node solveEnvVariables.js build/src/constants.js",
         "install-locally-dev": "npm run install-locally -- dev",
         "build-dev": "npx tsc && node solveEnvVariables.js build/src/constants.js -- dev",
-        "checkChangesInVariables": "./scripts/checkConstants",
-        "checkUnwantedLogs": "./scripts/checkUnwantedLogs",
+        "checkChangesInVariables": "node scripts/checkConstants.js",
+        "checkUnwantedLogs": "node scripts/checkUnwantedLogs.js",
```
costinsin, line 16: Nice fix

Virgil993: true, unnecessary

PR #728, opened 2024-01-17
Add a hidden command that bundles a class and generates a zip and and ast summary for it.

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature

## Description

In this PR, I've implement a hidden command (it does not appear in `genezio help`) that can be used to bundle and generate the AST for a class. This is useful if you want to deploy separately a class (e.g.: auth scenario) as part of an CI/CD.

It generates in an output folder the zip that contains the bundle and the AST of that class.

### DIFF
--- a/src/bundlers/utils.ts
+++ b/src/bundlers/utils.ts
@@ -1,0 +1,79 @@
+import log from "loglevel";
+import { ClassConfiguration } from "../models/projectConfiguration.js";
+import { fileExists } from "../utils/file.js";
+import { printAdaptiveLog } from "../utils/logging.js";
+import { BundlerInterface, BundlerOutput } from "./bundler.interface.js";
+import { TsRequiredDepsBundler } from "./node/typescriptRequiredDepsBundler.js";
+import { TypeCheckerBundler } from "./node/typeCheckerBundler.js";
+import { NodeJsBundler } from "./node/nodeJsBundler.js";
+import { NodeJsBinaryDependenciesBundler } from "./node/nodeJsBinaryDependenciesBundler.js";
+import { BundlerComposer } from "./bundlerComposer.js";
+import { DartBundler } from "./dart/dartBundler.js";
+import { KotlinBundler } from "./kotlin/kotlinBundler.js";
+import { debugLogger } from "../utils/logging.js";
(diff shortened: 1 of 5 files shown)


src/bundlers/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+import { NodeJsBinaryDependenciesBundler } from "./node/nodeJsBinaryDependenciesBundler.js";
+import { BundlerComposer } from "./bundlerComposer.js";
+import { DartBundler } from "./dart/dartBundler.js";
+import { KotlinBundler } from "./kotlin/kotlinBundler.js";
+import { debugLogger } from "../utils/logging.js";
+import { createTemporaryFolder } from "../utils/file.js";
+import { ProjectConfiguration } from "../models/projectConfiguration.js";
+import { Program } from "../models/genezioModels.js";
+
+export async function bundle(
```
costinsin, line 17: I have seen that this code is duplicated twice in `deploy.ts` line 289 and in `local.ts` line 508 (with small differences).

Can you please only write it here in `utils.ts` and reuse it everywhere (3 places). It would help if we want to change the bundling system in the future, because we would not need to change the code in 3 places.

src/commands/bundle.ts
```diff
@@ -0,0 +1,44 @@
...
+import { GenezioBundleOptions } from "../models/commandOptions.js";
+import { SdkGeneratorResponse } from "../models/sdkGeneratorResponse.js";
+import { getProjectConfiguration } from "../utils/configuration.js";
+import { ProjectConfiguration } from "../models/projectConfiguration.js";
+import { bundle } from "../bundlers/utils.js";
+import { mkdirSync } from "fs";
+import { writeToFile, zipDirectory } from "../utils/file.js";
+import path from "path";
+
+export async function bundleCommand(options: GenezioBundleOptions) {
```
costinsin, line 12: Same for this code. It's duplicated twice in `deploy.ts` line 258 and `generateSdk.ts` line 61. It would be nice to have 1 function instead of 3.

radu1122 (approved):

costinsin (approved):

PR #747, opened 2024-01-26
GNZ-1395 Include token in the GnzContext object if it is the first parameter

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature

## Description

If on the server side, the first argument is of type `GnzContext`, the SDK will try to populate it with the token from the local storage.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/generateSdk/sdkGenerator/JsSdkGenerator.ts
+++ b/src/generateSdk/sdkGenerator/JsSdkGenerator.ts
@@ -29,17 +29,74 @@ const template = `/**
 
 import { Remote } from "./remote.js"
 
+{{#hasGnzContext}}
+export class LocalStorageWrapper implements Storage {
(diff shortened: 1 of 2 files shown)


src/generateSdk/sdkGenerator/JsSdkGenerator.ts
```diff
@@ -29,18 +29,75 @@ const template = `/**
 
 import { Remote } from "./remote.js"
 
+{{#hasGnzContext}}
```
stefandarius, line 32: this would be added in every class that uses GnzContext, right? wouldn't it be better to add it to the `remote` file to avoid duplicated code? worst case the code won't be used.

stefandarius (approved):

PR #757, opened 2024-01-30
Gnz 1326 go support

stefandarius:
## Type of change

-   [x] 🍕 New feature

## Description

Support for GO as a backend language in genezio

## Checklist


src/commands/deploy.ts
```diff
@@ -72,9 +77,15 @@ export async function deployCommand(options: GenezioDeployOptions) {
...
     // 0.1.* of @genezio/types.
     const packageJsonPath = path.join(backendCwd, "package.json");
-    if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", REQUIRED_GENEZIO_TYPES_VERSION_RANGE) === false) {
+    if (
+        isDependencyVersionCompatible(
+            packageJsonPath,
+            "@genezio/types",
+            REQUIRED_GENEZIO_TYPES_VERSION_RANGE,
+        ) === false
+    ) {
```
costinsin, line 86: I know you did not add this, but this is only valid for ts/js projects. It actually writes an error in the terminal when deploying.

```
ERROR: Error while reading package.json file: Error: ENOENT: no such file or directory, open '/home/costin/genezio/my-projects/gotest/package.json'
```

Maybe we should extract this logic into a function and treat it based on the project language. This function is also called in `genezio local`, so it should also be modified there.

src/commands/deploy.ts
```diff
@@ -325,7 +336,11 @@ export async function deployClasses(
             const archivePath = path.join(archivePathTempFolder, `genezioDeploy.zip`);
 
             debugLogger.debug(`Zip the directory ${output.path}.`);
-            await zipDirectory(output.path, archivePath);
+            if (element.language === ".go") {
+                await zipFile(path.join(output.path, "bootstrap"), archivePath);
+            } else {
+                await zipDirectory(output.path, archivePath);
+            }
```
costinsin, line 343: I don't like these `if`s that depend on language here. There is also one with `.kt` above these lines. I don't think you should solve this now, but maybe think about a good solution on how to run code based on which language arrives here. We should remove the ts/js bias in our codebase.

src/generateSdk/astGenerator/GoAstGenerator.ts
```diff
@@ -0,0 +1,144 @@
...
+                switch (astNode.type) {
+                    case AstNodeType.StructLiteral:
+                        const structLiteral: StructLiteral = {
+                            type: AstNodeType.StructLiteral,
+                            name: astNode.name,
+                            path: astNode.path,
+                            typeLiteral: astNode.typeLiteral,
+                        };
+                        body.push(structLiteral);
+                        break;
```
costinsin, line 127: Unexpected lexical declaration in case block. - eslint (no-case-declarations)

```suggestion
                    case AstNodeType.StructLiteral: {
                        const structLiteral: StructLiteral = {
                            type: AstNodeType.StructLiteral,
                            name: astNode.name,
                            path: astNode.path,
                            typeLiteral: astNode.typeLiteral,
                        };
                        body.push(structLiteral);
                        break;
                    }
```

src/utils/file.ts
```diff
@@ -116,6 +116,21 @@ export async function zipDirectory(sourceDir: string, outPath: string): Promise<
...
+    const stream = fs.createWriteStream(outPath);
+
+    return new Promise((resolve, reject) => {
+        archive
+            .file(sourcePath, { name: path.basename(sourcePath) })
+            .on("error", (err) => reject(err))
+            .pipe(stream);
+
+        stream.on("close", () => resolve());
+        archive.finalize();
```
costinsin, line 130: Is there a possibility that the Promise is resolved before `archive.finalize()` is called?

I am not sure about this, maybe you can explain if I'm wrong.
```suggestion
        archive.finalize();
        stream.on("close", () => resolve()); 
```

costinsin (approved):
LGTM! I tested it and it works. Please resolve my small feedback if possible before merging.

src/commands/deploy.ts
vladiulianbogdan: The error is loggerDebug it won't appear in production.

src/utils/file.ts
vladiulianbogdan: Even if the promise resolves, the flow will continue and `archive.finalize()` will still be called.

src/bundlers/go/goMain.ts
```diff
@@ -0,0 +1,151 @@
...
+    } else {
+        eventBody := []byte(event.Body)
+        // Decode the request body into struct and check for errors
+        err := json.Unmarshal(eventBody, &body)
+        if err != nil {
+            errorResponse := sendError(err)
+            return errorResponse, nil
+        }
+        // Call the appropriate method
+        switch body.Method {
```
vladiulianbogdan, line 107: In TS and JS the method is "Class.method". We don't want to do this here?

src/commands/deploy.ts
costinsin: Even though is not shown in production, I still think we should only check this for ts/js environments.

PR #758, opened 2024-01-30
Check if the @genezio/types version used is older and throw an error.

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix

## Description

We need to check if the user is using an older version of @genezio/types because we migrated the decorators implemented in the @genezio/types package to the stage 3 implementation. Otherwise, the user will get an error at runtime. This check can be removed in the future once no one is using version 0.1.* of @genezio/types.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -82,7 +82,7 @@ export async function deployCommand(options: GenezioDeployOptions) {
     const packageJsonPath = path.join(backendCwd, "package.json");
     if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", "1.0.0") === false) {
         log.error(
-            `You are currently using an older version of @genezio/types, which is not compatible with this version of the genezio CLI. To resolve this, please update the @genezio/types package on your server using the following command: npm install @genezio/types@^1.0.0`
+            `You are currently using an older version of @genezio/types, which is not compatible with this version of the genezio CLI. To solve this, please update the @genezio/types package on your backend component using the following command: npm install @genezio/types@^1.0.0`
         );
         exit(1);
     }
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -53,6 +53,7 @@ import { interruptLocalProcesses } from "../utils/localInterrupt.js";
 import { Status } from "../requests/models.js";
 import { loginCommand } from "./login.js";
 import { bundle } from "../bundlers/utils.js";
+import { isDependencyVersionCompatible } from "../utils/dependencyChecker.js";
 
 export async function deployCommand(options: GenezioDeployOptions) {
     await interruptLocalProcesses();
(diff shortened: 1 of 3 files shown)


src/commands/local.ts
```diff
@@ -158,6 +159,20 @@ export async function startLocalEnvironment(options: GenezioLocalOptions) {
...
     const yamlProjectConfiguration = await getProjectConfiguration();
 
+    // We need to check if the user is using an older version of @genezio/types
+    // because we migrated the decorators implemented in the @genezio/types package to the stage 3 implementation.
+    // Otherwise, the user will get an error at runtime. This check can be removed in the future once no one is using version
+    // 0.1.* of @genezio/types.
+    const packageJsonPath = yamlProjectConfiguration.workspace ? 
+        path.join(yamlProjectConfiguration.workspace.backend, "package.json") :
+        path.join(process.cwd(), "package.json");
+    if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", "1.0.0") == false) {
```
costinsin, line 169:
```suggestion
    if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", "1.0.0") === false) {
```

src/commands/deploy.ts
```diff
@@ -74,6 +75,18 @@ export async function deployCommand(options: GenezioDeployOptions) {
     const backendCwd = configuration.workspace?.backend || process.cwd();
     const frontendCwd = configuration.workspace?.frontend || process.cwd();
 
+    // We need to check if the user is using an older version of @genezio/types
+    // because we migrated the decorators implemented in the @genezio/types package to the stage 3 implementation.
+    // Otherwise, the user will get an error at runtime. This check can be removed in the future once no one is using version
+    // 0.1.* of @genezio/types.
+    const packageJsonPath = path.join(backendCwd, "package.json");
+    if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", "1.0.0") == false) {
```
costinsin, line 83:
```suggestion
    if (isDependencyVersionCompatible(packageJsonPath, "@genezio/types", "1.0.0") === false) {
```

src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,34 @@
...
+    dependency: string,
+    version: string,
+): boolean | undefined {
+    try {
+        const packageJson = JSON.parse(fs.readFileSync(path, "utf8"));
+
+        const depVersion =
+            packageJson.dependencies[dependency] || packageJson.devDependencies[dependency];
+        if (!depVersion) {
+            debugLogger.error(`Dependency ${dependency} not found.`);
```
costinsin, line 25: Shouldn't this be indented? I think you forgot to run prettier.

costinsin (approved):

src/commands/deploy.ts
```diff
@@ -74,6 +75,18 @@ export async function deployCommand(options: GenezioDeployOptions) {
...
+        log.error(
+            `You are currently using an older version of @genezio/types, which is not compatible with this version of the genezio CLI. To resolve this, please update the @genezio/types package on your server using the following command: npm install @genezio/types@^1.0.0`
```
costinsin, line 85: I think the better wording here is "fix"/"solve"
```suggestion
            `You are currently using an older version of @genezio/types, which is not compatible with this version of the genezio CLI. To solve this, please update the @genezio/types package on your backend component using the following command: npm install @genezio/types@^1.0.0`
```

PR #778, opened 2024-02-06
Properly serialize and deserialize errors over the wire

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature

## Description

Up until now, we did not properly serialize the errors. We were just calling `error.toString()` but that does not include too much information. The proposal in this PR is to serialize Errors containing the following fields: message, code, info. This can be done in a multi language environment and satisfies most of the use cases.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/bundlers/node/lambdaHandlerGenerator.ts
+++ b/src/bundlers/node/lambdaHandlerGenerator.ts
@@ -171,7 +171,7 @@ if (!genezioClass) {
                 body = JSON.parse(event.body);
             } catch (error) {
                 return {
-                    statusCode: 500,
+                    statusCode: 400,
                     body: JSON.stringify({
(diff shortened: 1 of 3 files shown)


stefandarius (approved):

costinsin (approved):

costinsin:
Are you planning to add a GenezioError type in the sdk, so the user can easily access the fields of the serialized error?

vladiulianbogdan:
> Are you planning to add a GenezioError type in the sdk, so the user can easily access the fields of the serialized error?

Yea, still debating where is the right place. For now, I've added it in `@genezio/types` for the server to use it. The frontend can also install it for now.

andreia-oca (approved):
Looks good!

PR #781, opened 2024-02-09
generate sdk and bundle based on workspace path

stefandarius:
## Type of change

-   [x] 🐛 Bug Fix

## Description

SDK generation and bundling must be performed in the backend directory instead of the project root to provide the right context for language-specific tools such as the go parser.
We can achieve this by passing an optional path to the backend folder in the AST generator interface

## Checklist

### DIFF
--- a/src/generateSdk/astGeneratorHandler.ts
+++ b/src/generateSdk/astGeneratorHandler.ts
@@ -27,7 +27,6 @@ interface AstGeneratorPlugin {
 export async function generateAst(
     input: AstGeneratorInput,
     plugins: string[] | undefined,
-    backendPath?: string,
 ): Promise<AstGeneratorOutput> {
     const extension = path.extname(input.class.path).replace(".", "");
     let pluginsImported: AstGeneratorPlugin[] = [];
@@ -86,7 +85,7 @@ export async function generateAst(
 
     const astGeneratorClass = new plugin.AstGenerator();
 
-    return await astGeneratorClass.generateAst(input, backendPath).catch((err) => {
+    return await astGeneratorClass.generateAst(input).catch((err) => {
         debugLogger.log("An error has occurred", err);
         throw Object.assign(err, { path: input.class.path });
     });
--- a/src/generateSdk/astGeneratorHandler.ts
+++ b/src/generateSdk/astGeneratorHandler.ts
@@ -27,6 +27,7 @@ interface AstGeneratorPlugin {
 export async function generateAst(
     input: AstGeneratorInput,
     plugins: string[] | undefined,
+    backendPath?: string,
 ): Promise<AstGeneratorOutput> {
     const extension = path.extname(input.class.path).replace(".", "");
     let pluginsImported: AstGeneratorPlugin[] = [];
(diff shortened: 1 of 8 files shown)


src/generateSdk/sdkGenerator/TsSdkGenerator.ts
```diff
@@ -292,7 +292,10 @@ class SdkGenerator implements SdkGeneratorInterface {
 
                 // @ts-expect-error A refactor need to be performed here to avoid this error
                 const methodView: MethodViewType = {
-                    name: methodDefinition.name,
+                    // methods start with lowercase
+                    name:
+                        methodDefinition.name.charAt(0).toLowerCase() +
+                        methodDefinition.name.slice(1),
```
costinsin, line 298: Why is this change needed? What if i want to have functions that start with upper case? It's a ts convention to have lowercase named methods, but we should not impose it.

src/generateSdk/astGeneratorHandler.ts
```diff
@@ -27,6 +27,7 @@ interface AstGeneratorPlugin {
 export async function generateAst(
     input: AstGeneratorInput,
     plugins: string[] | undefined,
+    backendPath?: string,
```
costinsin, line 30: I would name it `rootPath` everywhere.
```suggestion
    rootPath?: string,
```

src/generateSdk/generateSdkApi.ts
```diff
@@ -43,6 +43,7 @@ export async function sdkGeneratorApiHandler(
         const astGeneratorOutput = await generateAst(
             input,
             projectConfiguration.plugins?.astGenerator,
+            projectConfiguration.workspace?.backend,
```
costinsin, line 46:
```suggestion
            /* rootPath= */ projectConfiguration.workspace?.backend,
```

src/generateSdk/astGeneratorHandler.ts
costinsin: Also, what if we include this path inside `AstGeneratorInput`? this way we could have this nice `AstGeneratorInput -> AstGeneratorOutput`, instead of `(AstGeneratorInput, path) -> AstGeneratorOutput`

vladiulianbogdan (approved):

src/generateSdk/sdkGenerator/TsSdkGenerator.ts
stefandarius: This is needed because all methods in Go will start with an uppercase letter.
The client code for our templates in ts/js expects methods starting with lowercase. We need this for a full-stack template with a go server and a ts client to work.

src/generateSdk/astGeneratorHandler.ts
stefandarius: i would find it a bit confusing to name it root since we also have the notion of the root of the mono repo project. i personally think of that as root, event though in the context of AST generation the "root of the action" is the backend folder.
i will try to see if I can add the value as a field in the `AstGeneratorInput` class

src/generateSdk/sdkGenerator/TsSdkGenerator.ts
stefandarius: @costinsin i added the root property in the AstGeneratorInput class. much cleaner now

costinsin (approved):

PR #808, opened 2024-02-21
Update genezio runtime handler path handling for HTTP calls

valighita:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description

Runtime call url changed from `/CallFunction?funcId=x&method=y` to `/funcId/className/methodName`. Update the packages handler accordingly.


MagoDelBlocco (approved):
LGTM

PR #813, opened 2024-02-21
Lambda Handler Typescript Timeout error handle

cristim67:
## Type of change

-   [X] 🐛 Bug Fix

## Description

Lambda timeout is not handled and we lose users on that side because it looks like an internal error

## Checklist

### DIFF
--- a/src/bundlers/node/lambdaHandlerGenerator.ts
+++ b/src/bundlers/node/lambdaHandlerGenerator.ts
@@ -239,35 +239,43 @@ if (!genezioClass) {
             });
 
             try {
-                const response = Promise.resolve(object[method](...(body.params || [])))
-                    .then((result) => {
-                        return {
-                            statusCode: 200,
-                            body: JSON.stringify({
-                                jsonrpc: "2.0",
-                                result: result,
-                                error: null,
-                                id: requestId,
-                            }),
-                            headers: { 'Content-Type': 'application/json', 'X-Powered-By': 'genezio' }
-                        };
-                    })
-                    .catch(async (err) => {
-                        console.error(err);
-                        await sendSentryError(err);
-                        return {
-                            statusCode: 200,
-                            body: JSON.stringify({
-                                jsonrpc: "2.0",
-                                error: prepareForSerialization(err),
-                                id: requestId,
-                            }),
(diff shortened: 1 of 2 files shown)


src/bundlers/node/lambdaHandlerGenerator.ts
```diff
@@ -1,6 +1,8 @@
 //const object = new handler.genezio[Object.keys(handler.genezio)[0]]();
 /* eslint-disable no-useless-escape */
+import getUser from "../../requests/getUser.js";
 
+const user = await getUser();
```
radu1122, line 5: this call should be made inside a function upper level and only send the timeout as a param to the `lambdaHandlerGenerator` function to keep it more dynamic

src/bundlers/node/lambdaHandlerGenerator.ts
```diff
@@ -239,35 +241,43 @@ if (!genezioClass) {
             });
 
             try {
-                const response = Promise.resolve(object[method](...(body.params || [])))
```
radu1122, line 242: looks like there are too many modifications than expected here

src/bundlers/node/lambdaHandlerGenerator.ts
```diff
@@ -239,7 +239,13 @@ if (!genezioClass) {
...
+                const timeoutPromise = new Promise((resolve,reject) => {
```
radu1122, line 242: this should be for all calls, http, jsonrpc and cron

radu1122: i would also add a log for timeout

src/bundlers/node/nodeJsBundler.ts
```diff
@@ -437,7 +438,7 @@ export class NodeJsBundler implements BundlerInterface {
             writeToFile(
                 temporaryFolder,
                 "index.mjs",
-                handlerGenerator(`"${input.configuration.name}"`),
+                handlerGenerator(`"${input.configuration.name}"`, await getUser().then(user => user?.subscriptionLimits.executionTime) ?? 30),
```
radu1122, line 441: a catch should be adde here

radu1122 (approved):

PR #826, opened 2024-02-26
Force lowercase headers feed into handler

MagoDelBlocco:
-   [X] 🐛 Bug Fix

## Description
Fix request headers passed on to the actual user code. Previously, they were left untouched in this handler, but they came altered from the golang client code, which formatted them in HTTP/1.X upper-case compliant manner. This could potentially conflict with how users might format their requests and expect headers.

### DIFF
--- a/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
+++ b/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
@@ -83,8 +83,14 @@ if (!genezioClass) {
             }
 
             const method = components[2];
+
+            const http2CompliantHeaders = {};
+            for (const header in event.headers) {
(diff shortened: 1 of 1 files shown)


costinsin (approved):
LGTM

andreia-oca (approved):
LGTM

MagoDelBlocco:
> > This could potentially conflict with how users might format their requests and expect headers.
> 
> Is this something that we need to write documentation about?
> 
> Up until now, the interaction with the webhooks was done via `GenezioHttpRequest`/`GenezioHttpResponse` and this agnostically treats headers as a json [1]. Should we also elaborate on case sensitive headers/http versions?
> 
> https://genezio.com/docs/features/http-methods-webhooks

Once we release the runtime for production, this should be something that is documented. Basically, people should all expect their headers to be HTTP/2 compliant.

PR #837, opened 2024-02-28
Add Go bundler for genezio-runtime

valighita:
## Type of change

-   [x] 🍕 New feature

## Description

Implement bundler and handler generator for genezio runtime. The go code is compiled as a plugin which can be loaded at runtime in another Go application.
The plugin exports a function to handle requests, which is called by the runtime go wrapper after loading the plugin.

Webhooks are not supported yet.


costinsin:
I don't really like that we are duplicating a lot of logic inside here. How different are those handlers? Can't we have a logic that generates the common part of the handler the same for both lambda and runtime, and then just modify what is different about them?

If we in the future want to patch something at the bundler, we need to modify it in 3 places (local, lambda, runtime). Do you guys think this is duable?

valighita:
> we

Yes, that would be the right solution. I haven't done this since this work already needed to be done for lambda and local, so I figured some refactoring is needed.

stefandarius (approved):

PR #841, opened 2024-02-29
Add a mechanism that lets us read either the original config or one filled with default values

costinsin:
## Type of change
-   [x] 🍕 New feature

## Description

We have 2 use-cases when interacting with the genezio.yaml config:
- Reading it for consuming the information

  In this case we want the configuration to be filled with default
  values if the user does not provide values for optional fields.

  Example: If region is not specified, it should default to `us-east-1`

- Reading the file just to modify one field and the write it back

  In this case we do want to read the original configuration (with no
  default filled values) so that we don't modify the user configuration
  entierly, only the necessary fields.

  Example:
  ```yaml
  name: foo
  frontend:
      path: .
  ```

  We want to write the `subdomain` field to `frontend`. If we use
  the normal `read()`, some values will be defaulted (region for example)

  `read()`:
  ```diff
    name: foo
  + region: us-east-1
    frontend:
      path: .
  +   subdomain: bar
  ```

  `read(/* fillDefaults= */ false)`:
  ```diff
    name: foo
    frontend:
      path: .
  +   subdomain: bar
  ```

## Checklist

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -126,13 +126,17 @@ export async function deployCommand(options: GenezioDeployOptions) {
             configuration.backend &&
             process.env["DISABLE_AB_TESTING"] !== "true"
         ) {
-            configuration.backend.cloudProvider = await performCloudProviderABTesting(
+            const yamlConfig = await configIOController.read(/* fillDefaults= */ false);
+            if (!yamlConfig.backend) {
+                throw new Error("No backend entry in genezio configuration file.");
+            }
+            yamlConfig.backend.cloudProvider = await performCloudProviderABTesting(
                 configuration.name,
                 configuration.region,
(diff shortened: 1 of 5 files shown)


costinsin:
Current dependencies on/for this PR:
* **#841** <a href="https://app.graphite.dev/github/pr/Genez-io/genezio/841?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `dev`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Genez-io/genezio/841?utm_source=stack-comment).

andreia-oca (approved):
Looks great! Amazing code, chapeau bas 🚀

PR #844, opened 2024-03-01
first version of test interface component served on localhost

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature
-   [x] 🧑‍💻 Improvement

## Description

A lot of users felt that the testing interface when running genezio local should be served on localhost, this PR allows us to start a local http server for handling the requests of different classes from the user's project and also serves the testing interface on localhost:port/test-interface/local?=port. This is made posible by using the test interface component from the npm registry which also contains a bundled up version of a react app which can be served statically using express.

## Checklist

-   [ ] I have added tests;


src/commands/local.ts
```diff
@@ -652,10 +652,23 @@ async function startServerHttp(
...
 ): Promise<http.Server> {
     const app = express();
+    const __filename = url.fileURLToPath(import.meta.url);
+    const __dirname = path.dirname(__filename);
     app.use(cors());
     app.use(bodyParser.raw({ type: () => true, limit: "6mb" }));
     app.use(genezioRequestParser);
 
+    // serve test interface built folder on localhost
+    const builtFolder = path.resolve(
```
costinsin, line 662:
```suggestion
    const buildFolder = path.resolve(
```

src/commands/local.ts
```diff
@@ -652,10 +652,23 @@ async function startServerHttp(
...
+        __dirname,
+        "../../../node_modules/genezio-test-interface-component/dist/build",
+    );
+    app.use(express.static(builtFolder));
+    app.get(`/test-interface/local/`, (req, res) => {
```
costinsin, line 667: Bogdan Ripa had some ideas on what path we should use. Maybe talk to him about this. This path is too long and does not mean anything to the user.

```suggestion
    app.get(`something else`, (req, res) => {
```

src/commands/local.ts
```diff
@@ -1005,7 +1018,10 @@ To change the server version, go to your ${colors.cyan(
         newVersion,
     );
 
-    log.info("\x1b[32m%s\x1b[0m", `Test your code at ${LOCAL_TEST_INTERFACE_URL}?port=${port}`);
+    log.info(
+        "\x1b[32m%s\x1b[0m",
```
costinsin, line 1022: Can you please use `colors` instead?

src/commands/local.ts
```diff
@@ -652,10 +652,23 @@ async function startServerHttp(
...
     app.use(genezioRequestParser);
 
+    // serve test interface built folder on localhost
+    const builtFolder = path.resolve(
+        __dirname,
+        "../../../node_modules/genezio-test-interface-component/dist/build",
+    );
+    app.use(express.static(builtFolder));
+    app.get(`/test-interface/local/`, (req, res) => {
+        const filePath = path.join(builtFolder, "index.html");
```
costinsin, line 668: Does this work? Does it know how to also send the JS/CSS files?

src/commands/local.ts
```diff
@@ -60,7 +60,7 @@ import inquirer, { Answers } from "inquirer";
 import { DEFAULT_NODE_RUNTIME } from "../models/nodeRuntime.js";
 import { getNodeModulePackageJsonLocal } from "../generateSdk/templates/packageJson.js";
 import { compileSdk } from "../generateSdk/utils/compileSdk.js";
-import { exit } from "process";
+import { cwd, exit } from "process";
```
costinsin, line 63: Looks like you are not using it.

```suggestion
import { exit } from "process";
```

src/commands/local.ts
radu1122: we will go with `/explore`

PR #850, opened 2024-03-04
add eslint unusedvars ignore pattern

stefandarius:
## Description

make eslint ignore unused function parameters prefixed with `_`


costinsin (approved):
I liked the `no-console` rule

PR #856, opened 2024-03-04
Create a new axios instance to do ab testing without throwing errors

andreia-oca:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix

## Description

The AB Testing PR introduced a verbose error when the project was not found.
This behaviour is expected for projects that were not yet deployed at all. Hence, I want to fail silently.

To do that I created a new axios instance that does not use an interceptor. Unfortunately the code is mostly duplicated in the AB testing method, but this part of the codebase will be deleted after the ab testing is concluded.

## Checklist

### DIFF
--- a/src/utils/abTesting.ts
+++ b/src/utils/abTesting.ts
@@ -1,11 +1,36 @@
+import axios, { AxiosResponse } from "axios";
+import version from "../utils/version.js";
 import { CloudProviderIdentifier } from "../models/cloudProviderIdentifier.js";
-import getProjectInfoByName from "../requests/getProjectInfoByName.js";
+import { ProjectDetails, StatusOk } from "../requests/models.js";
+import { getAuthToken } from "./accounts.js";
+import { BACKEND_ENDPOINT } from "../constants.js";
+import { GENEZIO_NOT_AUTH_ERROR_MSG } from "../errors.js";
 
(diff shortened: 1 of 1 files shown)


src/utils/abTesting.ts
```diff
@@ -1,12 +1,33 @@
...
+                    },
+                    headers: {
+                        Authorization: `Bearer ${authToken}`,
+                        "Accept-Version": `genezio-cli/${version}`,
+                    },
+                });
+            if (response.data.project) {
+                return true;
+            }
         }
```
costinsin, line 31: Prefer early returns because they don't add indentation

```suggestion
        if (authToken) {
            return false;
        }

        const response: AxiosResponse<StatusOk<{ project: ProjectDetails }>> =
            await uninterceptedAxiosInstance({
                method: "GET",
                url: `${BACKEND_ENDPOINT}/projects/name/${name}`,
                params: {
                    region: region,
                },
                headers: {
                    Authorization: `Bearer ${authToken}`,
                    "Accept-Version": `genezio-cli/${version}`,
                },
            });

        if (response.data.project) {
            return true;
        }
```

costinsin (approved):

costinsin:
I don't really like this idea of duplicating code, but I don't have a better solution for now.

PR #860, opened 2024-03-05
[WIP] Gnz 628 error handling

stefandarius:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description

This PR aims to improve the CLI experience for users and developers and declare a standard error-handling mechanism that should be used throughout the project.

Notable changes:
- the logging module has been changed from `loglevel` to `tslog`, which uses the same API but more colorful and configurable.
- there are 2 logger instances available in the CLI: a `debugLogger` which is by default disabled for production and enabled for development, and a `mainLogger` (exported as `log`), which should be used to display any messages for the end user
- each command is responsible for catching errors at the top level, displaying them using the `logError` function defined in the `logging` file, and terminating the execution of the process. No other functions should display errors using the `mainLogger`, or terminate the execution of the process.

## Checklist


src/commands/deploy.ts
```diff
@@ -219,17 +219,17 @@ export async function deployCommand(options: GenezioDeployOptions) {
                 options,
             ).catch(async (error) => {
                 if (error instanceof Error) {
-                    log.error(error.message);
                     if (error.message == "No frontend entry in genezio configuration file.") {
+                        log.error(error.message);
```
vladiulianbogdan, line 223: We are still doing `exit(0)` here. Is this indended?

stefandarius: there are still a few places where I need to enforce the new error handling mechanism, as well as integrate it with the Axios interceptor. I am working on this right now.

costinsin (approved):

PR #906, opened 2024-03-19
Gnz 1479 enforce return type checking for webhooks

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
-   [x] 🔥 Breaking change
-   [x] 📝 Documentation Update

## Description

This PR aims to enforce the types used as parameters and as return types in Http Methods. We have decided to change the way we quide our users when working with WebHooks. We now aim to encourage the usage of the native node types for Request and for Response. As such, before generating the SDK, we check if the Http methods have the right configuration and if not, we don't continue the SDK generation

## Checklist

-   [ ] I have added tests;


src/generateSdk/astGenerator/TsAstGenerator.ts
```diff
@@ -142,9 +144,7 @@ export class AstGenerator implements AstGeneratorInterface {
                     }
                 }
                 if (!typeAtLocationPath) {
-                    throw new UserError(
-                        `Type ${escapedText} is not supported by genezio. Take a look at the documentation to see the supported types. https://docs.genezio.com/`,
-                    );
+                    return { type: AstNodeType.UnsupportedType, name: escapedText };
```
stefandarius, line 147: I would perform an additional check here to see if the type is indeed a Request or Response type (should be an interface defined in @types/node), and if so, Map it to a `AstNodeType.RequestType/ResponseType` (you should also define those in `genezioModels`).
This way the check you perform later in the SDK generation would be much simpler and language-agnostic.

Take a look at this [example](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#type-checker-apis) on how to use the TypeChecker to get the type of a symbol.

src/generateSdk/generateSdkApi.ts
```diff
@@ -55,12 +62,29 @@ export async function sdkGeneratorApiHandler(
...
+                    node.type === AstNodeType.ClassDefinition &&
+                    (node as ClassDefinition).name === classConfiguration.name,
+            ) as ClassDefinition;
+            const currentInvalidHttpMethods = checkInvalidHttpMethods(
+                classConfiguration,
+                astClassDefinition,
+            );
+            invalidHttpMethods = invalidHttpMethods.concat(currentInvalidHttpMethods.methods);
+            invalidHttpMethodsString += currentInvalidHttpMethods.methodsString;
+        }
```
stefandarius, line 77: As we discussed yesterday, I would perform this check in the bundling step. The SDK generation may not even happed if the user chooses not to, but the code still has to be correct.

stefandarius (changes requested):

src/generateSdk/generateSdkApi.ts
```diff
@@ -55,12 +62,30 @@ export async function sdkGeneratorApiHandler(
...
+            ) as ClassDefinition;
+            const currentInvalidHttpMethods = checkInvalidHttpMethods(
+                classConfiguration,
+                astClassDefinition,
+                language,
+            );
+            invalidHttpMethods = invalidHttpMethods.concat(currentInvalidHttpMethods.methods);
+            invalidHttpMethodsString += currentInvalidHttpMethods.methodsString;
+        }
+
```
stefandarius, line 79: This makes the function a bit hard to read. I propose to refactor the `checkInvalidHttpMethods` function to integrate this logic and check the HTTP method for that class, and also throw a UserError if it finds any problems. I think it Is ok to throw there and show the errors for just one class at a time.
If the function finds no problems, it should just return void and continue execution.

src/generateSdk/astGenerator/TsAstGenerator.ts
```diff
@@ -28,6 +28,9 @@ import {
     EnumType,
     DateType,
     MapType,
+    UnsupportedType,
```
stefandarius, line 31: Remove this type since it is not used anymore

src/generateSdk/astGenerator/TsAstGenerator.ts
```diff
@@ -142,6 +148,12 @@ export class AstGenerator implements AstGeneratorInterface {
                     }
                 }
                 if (!typeAtLocationPath) {
+                    if (escapedText == "Request") {
+                        return { type: AstNodeType.RequestType, name: escapedText };
+                    }
+                    if (escapedText == "Response") {
+                        return { type: AstNodeType.ResponseType, name: escapedText };
+                    }
```
stefandarius, line 156: I don't think this is enough. The user could define an interface called Request which would satisfy this check, but result in a runtime error. We should use the type checker to determine if that type really is the interface defined in @types/node

src/generateSdk/generateSdkApi.ts
```diff
@@ -74,6 +99,60 @@ export async function sdkGeneratorApiHandler(
...
+                            astMethod.params[0].paramType.name != "Request" ||
+                            astMethod.params[0].optional != false ||
+                            astMethod.returnType.type != AstNodeType.PromiseType ||
+                            astMethod.returnType.generic.type != AstNodeType.ResponseType ||
+                            (astMethod.returnType.generic as ResponseType).name != "Response")
+                    ) {
+                        methods.push(astMethod);
+                        methodsString +=
+                            ` ${colors.red(`- ${astClass.name}.${astMethod.name}`)}` + "\n";
+                    }
```
stefandarius, line 149: This is a lot of duplicated code here. This check should not be language-dependent, as we have the AST. Simply checking if the function has a single parameter of type `Request` and a return type of either `Response` or `Promise` with `generic = Response` should do the trick. There is no need to verify the names of the types here. The AST already did this

stefandarius (changes requested):

src/generateSdk/generateSdkApi.ts
```diff
@@ -74,6 +87,30 @@ export async function sdkGeneratorApiHandler(
...
+    let methodsString = "";
+    for (const method of sdkClass.methods) {
+        if (method.type === TriggerType.http) {
+            const astMethod = astClass.methods.find((m) => m.name === method.name);
+            if (
+                astMethod &&
+                (astMethod.params.length != 1 ||
+                    astMethod.params[0].paramType.type != AstNodeType.RequestType ||
+                    astMethod.params[0].optional != false ||
+                    astMethod.returnType.type != AstNodeType.PromiseType ||
```
stefandarius, line 101: I would also add the possibility for the return type to be directly `ResponseType`, to make it generic for other languages that will never return promises.

stefandarius (approved):

src/errors.ts
```diff
@@ -13,6 +14,10 @@ export const GENEZIO_NOT_AUTH_ERROR_MSG =
 export const PORT_ALREADY_USED = function (port: number) {
     return `The port ${port} is already in use. Please use a different port by specifying --port <port> to start your local server.`;
 };
+
+export const INVALID_HTTP_METHODS_FOUND = function (invalidHttpMethodsString: string) {
+    return `Invalid http methods found, the following methods are invalid:\n${invalidHttpMethodsString}\n${colors.yellow(`Please check the documentation for more information about how to define http methods\nhttps://genezio.com/docs/features/http-methods-webhooks`)}`;
```
radu1122, line 19: this sound me more explicit that the return value of the HTTP method is not correct

PR #915, opened 2024-03-25
parse go decorators with treesitter

stefandarius:
## Type of change

-   [x] 🍕 New feature

## Description

Use tresitter to enable users to specify deployable go services with comment annotation like `// genezio: deploy`, instead of relying on the yaml configuration file.

Current available annotations:
- genezio: deploy (above a struct with methods)
- genezio: http (above a method)
- genezio: cron <cron-string> (above a method)

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/utils/configuration.ts
+++ b/src/utils/configuration.ts
@@ -1,206 +1,26 @@
 import path from "path";
-import { createRequire } from "module";
 import { YAMLBackend, YamlClass, YamlMethod } from "../yamlProjectConfiguration/v2.js";
 import { getAllFilesFromPath } from "./file.js";
-import babel from "@babel/core";
-import fs from "fs";
-import FileDetails from "../models/fileDetails.js";
 import { debugLogger } from "./logging.js";
-import { NodePath } from "@babel/traverse";
-import {
-    ClassDeclaration,
-    ClassMethod,
-    isCallExpression,
-    isIdentifier,
-    isClassDeclaration,
-    CallExpression,
-    isObjectExpression,
-    isObjectProperty,
-    isStringLiteral,
-} from "@babel/types";
 import { TriggerType } from "../yamlProjectConfiguration/models.js";
 import { UserError } from "../errors.js";
-
-type MethodDecoratorInfo = {
-    name: string;
-    arguments?: { [key: string]: string };
-};
-
-type MethodInfo = {
-    name: string;
-    decorators: MethodDecoratorInfo[];
-};
-
-type ClassDecoratorInfo = {
-    name: string;
-    arguments?: { [key: string]: string };
-};
-
-type ClassInfo = {
-    path: string;
-    name: string;
-    decorators: ClassDecoratorInfo[];
-    methods: MethodInfo[];
-};
-
-function parseArguments(args: CallExpression["arguments"]): { [key: string]: string } {
-    if (!args) {
-        return {};
-    }
-
-    return args
-        .map((arg) => {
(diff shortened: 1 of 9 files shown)


src/yamlProjectConfiguration/migration.ts
```diff
@@ -56,6 +57,12 @@ export async function tryV2Migration(config: unknown): Promise<v2 | undefined> {
         } else {
             const scannedClasses = await scanClassesForDecorators({
                 path: v1Config.workspace?.backend || ".",
+                language: {
+                    name: Language.ts,
+                    runtime: "nodejs20.x",
+                    packageManager: PackageManagerType.npm,
+                },
+                classes: [],
```
costinsin, line 65: `classes` is not necessary

```suggestion
```

src/yamlProjectConfiguration/migration.ts
```diff
@@ -56,6 +57,12 @@ export async function tryV2Migration(config: unknown): Promise<v2 | undefined> {
         } else {
             const scannedClasses = await scanClassesForDecorators({
                 path: v1Config.workspace?.backend || ".",
+                language: {
+                    name: Language.ts,
+                    runtime: "nodejs20.x",
+                    packageManager: PackageManagerType.npm,
```
costinsin, line 63: these are not necessary either.

```suggestion
```

costinsin (approved):

src/yamlProjectConfiguration/migration.ts
stefandarius: I actually get this error if I try to remove those `Property 'packageManager' is missing in type '{ name: Language.ts; }' but required in type '{ packageManager: NonNullable<PackageManagerType | undefined>; }'. `

stefandarius:
> You also need to update the documentation.

I will, currently I am trying to solve the problem with the unit tests. Seems to be something with the way the tree-sitter package is built :(

src/utils/configuration.ts
```diff
@@ -1,206 +1,26 @@
...
+        decoratorExtractor.fileFilter(cwd),
+    );
 
     return await Promise.all(
-        allJsFilesPaths.map((file) => {
+        allFilesPaths.map((file) => {
             const filePath = path.join(cwd, file.path);
-            return getDecoratorsFromFile(filePath).catch((error) => {
+            return decoratorExtractor.getDecoratorsFromFile(filePath).catch((error) => {
                 if (error.reasonCode == "MissingOneOfPlugins") {
```
vladiulianbogdan, line 24: This if is specific to ts / js decorators. Maybe move it inside the `jsTsDecorator.ts`?

src/utils/decorators/jsTsDecorators.ts
```diff
@@ -0,0 +1,304 @@
...
+                (file.extension === ".js" || file.extension === ".ts") &&
+                !file.path.includes("node_modules") &&
+                !file.path.includes(".git") &&
+                !fs.lstatSync(folderPath).isDirectory()
+            );
+        };
+    }
+
+    async getDecoratorsFromFile(file: string): Promise<ClassInfo[]> {
+        const inputCode = fs.readFileSync(file, "utf8");
```
vladiulianbogdan, line 36: maybe make a private function from all this code and name it more specific: `getDecoratorsUsingBabel`. And then `getDecoratorsFromFile` will call the two methods.

vladiulianbogdan (approved):

PR #923, opened 2024-03-27
Add support for x86 binary dependencies

MagoDelBlocco:
The binary dependency bundler was adding arm64 binary dependencies, even if your runtime and k8s infrastructure are running on x86.

### DIFF
--- a/src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
+++ b/src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
@@ -12,8 +12,8 @@ import { UserError } from "../../errors.js";
 import { $ } from "execa";
 
 enum Architecture {
-    ARM = "arm64",
-    X86 = "amd64",
+    ARM64 = "arm64",
+    X64 = "x64",
 }
 
 export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
@@ -102,8 +102,8 @@ export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
         );
         const architecture =
             input.projectConfiguration.cloudProvider == CloudProviderIdentifier.GENEZIO
-                ? Architecture.ARM
-                : Architecture.X86;
+                ? Architecture.ARM64
+                : Architecture.X64;
         // 4. Redownload binary dependencies if necessary
         await this.#handleBinaryDependencies(
             input.extra.dependenciesInfo,
--- a/src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
+++ b/src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
@@ -1,17 +1,27 @@
 import path from "path";
 import fs from "fs";
 import { BundlerInput, BundlerInterface, BundlerOutput, Dependency } from "../bundler.interface.js";
+import { CloudProviderIdentifier } from "../../models/cloudProviderIdentifier.js";
 import { fileExists } from "../../utils/file.js";
 import { log } from "../../utils/logging.js";
 import { debugLogger } from "../../utils/logging.js";
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import exec from "await-exec";
 import packageManager from "../../packageManagers/packageManager.js";
 import { UserError } from "../../errors.js";
+import { $ } from "execa";
+
+enum Architecture {
+    ARM = "arm64",
+    X86 = "amd64",
+}
 
 export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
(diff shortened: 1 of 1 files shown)


src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
```diff
@@ -11,7 +12,11 @@ import packageManager from "../../packageManagers/packageManager.js";
 import { UserError } from "../../errors.js";
 
 export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
-    async #handleBinaryDependencies(dependenciesInfo: Dependency[], tempFolderPath: string) {
+    async #handleBinaryDependencies(
+        dependenciesInfo: Dependency[],
+        tempFolderPath: string,
+        architecture: string,
```
costinsin, line 18: Can you make this an enum? This way we ensure no other weird strings can be passed.

I'm thinking something like:
```ts
enum Architecture {
    ARM;
    X86;
}
```

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
```diff
@@ -66,11 +71,13 @@ export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
...
             try {
-                const { stdout, stderr } = await exec(
-                    "npx node-pre-gyp --update-binary --fallback-to-build --target_arch=arm64 --target_platform=linux --target_libc=glibc clean install " +
-                        dependency.name,
-                    { cwd: dependency.path },
-                );
+                const baseCommand =
+                    architecture == "arm"
+                        ? "npx node-pre-gyp --update-binary --fallback-to-build --target_arch=arm64 --target_platform=linux --target_libc=glibc clean install "
+                        : "npx node-pre-gyp --update-binary --fallback-to-build --target_arch=amd64 --target_platform=linux --target_libc=glibc clean install ";
```
costinsin, line 77: Maybe a switch?

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
```diff
@@ -66,11 +71,13 @@ export class NodeJsBinaryDependenciesBundler implements BundlerInterface {
...
+                const { stdout, stderr } = await exec(baseCommand + dependency.name, {
```
costinsin, line 78: Can you please replace this old `exec` with the new `execa` library? We are trying to make a slow transition when possible to `execa` because it's designed to be cross platform. Here is an example:
https://github.com/Genez-io/genezio/blob/main/src/packageManagers/npm.ts#L9

costinsin (approved):
LGTM

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
MagoDelBlocco: Managed to get around this by using the `execa` syntax.

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
MagoDelBlocco: Done, looks better, thanks!

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
MagoDelBlocco: Done, thank you for the suggestion, its much better now!

andreia-oca (approved):
Looks good to me!

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
```diff
@@ -1,17 +1,27 @@
...
 import { debugLogger } from "../../utils/logging.js";
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import exec from "await-exec";
 import packageManager from "../../packageManagers/packageManager.js";
 import { UserError } from "../../errors.js";
+import { $ } from "execa";
+
+enum Architecture {
+    ARM = "arm64",
```
valighita, line 15: Q: Why not ARM64 and AMD64?

valighita (approved):
LGTM

src/bundlers/node/nodeJsBinaryDependenciesBundler.ts
MagoDelBlocco: I feel like x86 is a more common name for the type of architecture that we are referring to.

valighita: But both `arm` and `x86` are different architectures from `arm64` and `amd64`. You can compile for those as well using node-gyp

MagoDelBlocco: Done.

PR #943, opened 2024-04-02
Fix error handling in genezio-remote.

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix

## Description

We response status was != 200, we would reject from genezio remote instead of just returning the JSON received.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/sdk-external-modules/remote-ts/src/remote.ts
+++ b/sdk-external-modules/remote-ts/src/remote.ts
@@ -12,10 +12,6 @@ async function makeRequestBrowser(request: any, url: any) {
         body: JSON.stringify(request),
     });
 
-    if (!response.ok) {
-        return response.json().then((error) => Promise.reject(error));
-    }
-
     return response.json();
 }
 


sdk-external-modules/remote-ts/src/remote.ts
```diff
@@ -12,10 +12,6 @@ async function makeRequestBrowser(request: any, url: any) {
         body: JSON.stringify(request),
     });
 
-    if (!response.ok) {
-        return response.json().then((error) => Promise.reject(error));
```
stefandarius, line 16: Isn't this part of the mechanism we agreed on a while ago to use native error handling in the SDK? Meaning that a failed request should result in the SDK throwing an error that the user should take care of?

vladiulianbogdan: Yes, but this happens at a higher level in the class SDK. Here, we threw the object which contains both `jsonrpc: 2.0` instead of just returning the error property.

In the class SDK, we do something like this (pseudocode):

```
if (response.error) {
   throw response.error
}

So, to summarise, this is a bug that blocked us from sending the proper error object to the user.


```

PR #952, opened 2024-04-04
Test auth stages on local

Virgil993:
# Pull Request Template

## Type of change

-   [ ] 🐛 Bug Fix
-   [ ] 🧑‍💻 Improvement

## Description

This PR adds the `--stage` option for `genezio local` which allows the user to test the Auth Service from different environments on local. Still need to add documentation

## Checklist

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -394,7 +394,11 @@ async function startProcesses(
     });
 
     const bundlersOutput = await Promise.all(bundlersOutputPromise);
-    await importServiceEnvVariables(projectConfiguration.name, projectConfiguration.region);
+    await importServiceEnvVariables(
+        projectConfiguration.name,
+        projectConfiguration.region,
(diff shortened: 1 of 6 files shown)


src/commands/local.ts
```diff
@@ -394,7 +395,24 @@ async function startProcesses(
     });
 
     const bundlersOutput = await Promise.all(bundlersOutputPromise);
-    await importServiceEnvVariables(projectConfiguration.name, projectConfiguration.region);
+
+    const authToken = await getAuthToken();
+    if (!authToken) {
+        process.env["LOGGED_IN_LOCAL"] = "false";
```
vladiulianbogdan, line 401: Two observations:

- I would set the env variable only if the user is authenticated.
- I would name the variable something with the prefix `GNZ_` or `GENEZIO_` to avoid the chance of clashing with user's envs.

vladiulianbogdan (approved):

PR #962, opened 2024-04-10
added try catch on delete folder

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
## Description
If you would terminate the proccess in the middle of a deploy, there was a chance that you enter a state where the program can't delete the temporary folder because most likely it is being written in at the same time. This would lead to a state where the proccess would not stop no matter how many times you try to terminate it. To fix this, i added a try catch on delete folder that logs the error, behaivour which is normal for mid execution interupts for any other platform.

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ import program from "./genezio.js";
 import { GenezioTelemetry, TelemetryEventTypes } from "./telemetry/telemetry.js";
 import { deleteFolder } from "./utils/file.js";
 import { SENTRY_DSN } from "./constants.js";
-import { debugLogger, logError } from "./utils/logging.js";
+import { debugLogger } from "./utils/logging.js";
 import path from "path";
 import os from "os";
 
@@ -40,7 +40,7 @@ process.on("SIGINT", async () => {
     try {
         await deleteFolder(temporaryFolder);
     } catch (error) {
-        logError(error as Error);
+        debugLogger.error("Error deleting temporary folder", error);
     }
     process.exit();
 });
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -37,7 +37,11 @@ process.on("SIGINT", async () => {
         commandOptions: "",
     });
 
-    await deleteFolder(temporaryFolder);
+    try {
+        await deleteFolder(temporaryFolder);
+    } catch (error) {
+        logError(error as Error);
```
costinsin, line 43: I am not sure but I think you should convert this to a UserError, else it will show up in our Sentry.

stefandarius: It won't show up in sentry because the error is not thrown anymore. It is just logged

Virgil993: I was thinking the same thing but the problem is  it is not a UserError because it happens when the excution is inturupted midway, the User doesn't need to do anything to fix this error. I was thinking that on any other framework, if you stop the execution mid way and you get one such error you get the full stack. The other option would be to not log anything. If you think that's a better option i will change it

stefandarius:
From my understanding, this fix will just prevent the error from being reported to Sentry, but the user will see the same message, which is pretty ugly IMO. Apart from that, temporary folders will still be on the disk.

If we can't solve this issue, meaning that temporary folders should be deleted at the end of the process no matter what, we should at least provide a more friendly error message for this scenario (or not log anything at all? Technically genezio did not crash, just the cleanup)

src/index.ts
costinsin: @stefandarius the sentry error will still appear because that function sends an event when the error has type Error.

Given the fact that this code is a process cleanup and the user can't influence it, I would consider silencing it (if it fails don't show anything).

@Virgil993 what do you think?

stefandarius: You're right. I wrote that function, but at a quick glance I thought it was a regular `log.error`.
I vote for silencing it too.

Virgil993: But i still think we should show a debug log

costinsin:
Please use .catch() method if you have only one statement in the try catch block and that statement is a promise. It is much more readable.

costinsin (approved):

PR #965, opened 2024-04-10
fix cluster deployments http methods

adriangeorge:
-   [x] 🐛 Bug Fix

## Description
Fixes http method calls on genezio cluster deployments


andreia-oca (approved):
LGTM!

PR #966, opened 2024-04-11
catch internal server errors and display appropiate message

stefandarius:
## Type of change

-   [x] 🐛 Bug Fix

## Description

This addresses the popular `error reading property code of undefined`, which occurs whenever the server crashes for various reasons and the HTTP response doesn't have a structure that can be understood by the CLI. 

This change catches such errors and encourages the users to open a GitHub issue regarding their problem.

## Checklist


costinsin (approved):
LGTM

PR #991, opened 2024-04-17
Remove libnpmpack and run npm pack instead

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

The `libnpmpack` dependency was to heavy and it was causing installation problems. We a

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/generateSdk/sdkWriter/jsSdkWriter.ts
+++ b/src/generateSdk/sdkWriter/jsSdkWriter.ts
@@ -33,7 +33,10 @@ async function writeSdk(
         await compileSdk(sdkPath, packageJson, language, publish);
     } else {
         if (exportAsTarball) {
-            const absoluteOutputPath = path.resolve(outputPath!);
+            if (!outputPath) {
+                throw new Error("Output path is required when exporting as a tarball.");
+            }
+            const absoluteOutputPath = path.resolve(outputPath);
             if (!fs.existsSync(absoluteOutputPath)) {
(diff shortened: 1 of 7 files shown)


src/generateSdk/sdkWriter/jsSdkWriter.ts
```diff
@@ -33,9 +33,12 @@ async function writeSdk(
         await compileSdk(sdkPath, packageJson, language, publish);
     } else {
         if (exportAsTarball) {
+            const absoluteOutputPath = path.resolve(outputPath!);
```
costinsin, line 36: What guarantees the output path will not be `undefined`?

costinsin: Why don't you change the function signature instead if you want to force `outputPath` to always be defined?

vladiulianbogdan: This method is called from multiple places: genezio sdk, genezio local, genezio deploy. This branch with tarball is currently used only by `genezio sdk` and it has a default value on `outputPath`. 

In case this path will be used also in other situations, I've add a check that throws an error.

costinsin (approved):
LGTM

PR #1003, opened 2024-04-19
Create rate limiter middleware

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature

## Description
I added a check on the Lambda handler generator that populates a GenezioContext object with the requestContext and the headrs of the incoming request from AWS. I also added a check in the Ts/Js SdkGenerator to add a new property to a GnzContext object named `isGnzContext`. this property is added only if the first parameter of a function is a GnzContext. It helps the lambdaHandler to know whether to populate that parameter or not. 

## Checklist

### DIFF
--- a/src/bundlers/node/lambdaHandlerGenerator.ts
+++ b/src/bundlers/node/lambdaHandlerGenerator.ts
@@ -76,6 +76,8 @@ if (!genezioClass) {
     }
 
     handler = handler ?? async function (event, context) {
+        console.log("DEBUG: event: " + JSON.stringify(event));
+        console.log("DEBUG: context: " + JSON.stringify(context));
         if (event.genezioEventType === "cron") {
             const method = event.methodName;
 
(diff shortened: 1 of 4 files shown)


stefandarius (approved):

src/bundlers/node/lambdaHandlerGenerator.ts
```diff
@@ -238,6 +238,10 @@ if (!genezioClass) {
                 });
             });
 
+            if(body.params && body.params.length > 0 && body.params[0].isGnzContext === true ) {
```
vladiulianbogdan, line 241: We have a handler for runtime also. Should we put this also there?

vladiulianbogdan (approved):

src/bundlers/node/lambdaHandlerGenerator.ts
Virgil993: im not sure how the event coming from runtime looks like and i don't know if it's gonna change in the future so i don't know if we should put it just yet

PR #1026, opened 2024-04-26
Fix connection error fallback mechanism

stefandarius:
## Type of change

-   [x] 🐛 Bug Fix

## Description

We were receiving sentry errors regarding connection to the local server for users' applications. This is an expected error that used to have a fallback mechanism that would attempt to restart the process. That mechanism was lost because we stopped returning AxiosErrors from the Axios interceptor.

This PR returns the unmodified AxiosError from the interceptor in case of a missing response (indicating a problem reaching the server), which will allow the fallback to take place, as well as sending a custom UserError which should no longer be captured by Sentry.

## Checklist

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -626,13 +626,28 @@ async function startServerHttp(
             return;
         }
 
-        const response = await communicateWithProcess(
-            localProcess,
-            req.params.className,
-            reqToFunction,
-            processForClasses,
-        );
-        sendResponse(res, response.data);
+        try {
+            const response = await communicateWithProcess(
+                localProcess,
+                req.params.className,
+                reqToFunction,
+                processForClasses,
+            );
+            sendResponse(res, response.data);
+        } catch (error) {
+            sendResponse(res, {
+                body: JSON.stringify({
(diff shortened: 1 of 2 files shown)


costinsin (commented):
What happens if the error is thrown inside local? It seems like the user will see the connection error even if the processes are now restarted. It might be confusing.

src/commands/local.ts
```diff
@@ -1018,6 +1018,7 @@ async function communicateWithProcess(
                 processForClasses,
                 localProcess.envVars,
             );
+            throw new UserError("There was an error connecting to the server");
```
costinsin, line 1021: I would expect that a retry mechanism would not throw error immediately after it tried to recover from error.

stefandarius: You are right. After testing it by manually killing the process associated with a class, it did restart the class, but the error was actually lost, it never showed. Since the initial request will fail, and after that we will restart the class, we should at least log an informative message that the request failed but things should work from now. Will include this in my next commit

costinsin (approved):

PR #1044, opened 2024-05-09
Move constructor within call

MagoDelBlocco:
-   [x] 🧑‍💻 Improvement

### DIFF
--- a/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
+++ b/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
@@ -35,7 +35,7 @@ if (!genezioClass) {
 
     let object;
 
-    handler = handler ?? async function (event) {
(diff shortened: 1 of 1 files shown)


src/bundlers/node/genezioRuntimeHandlerGenerator.ts
```diff
@@ -63,6 +44,26 @@ if (!genezioClass) {
             return response;
         }
 
+        try {
+            object = object ?? new genezioClass();
```
valighita, line 48: This can be written as `object ??= new genezioClass();`

valighita (approved):
LGTM

src/bundlers/node/genezioRuntimeHandlerGenerator.ts
MagoDelBlocco: Thanks!

PR #1045, opened 2024-05-09
Remove a/b testing for runtime

MagoDelBlocco:
Remove a/b testing from deployments. I've decided to leave the a/b testing logic there in case we will need it on a future cloud provider (genezio-unikernel is the first one to come to mind).

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -3,7 +3,6 @@ import { log } from "../utils/logging.js";
 import path from "path";
 import { exit } from "process";
 import {
-    ENVIRONMENT,
     DASHBOARD_URL,
     RECOMMENTDED_GENEZIO_TYPES_VERSION_RANGE,
     REQUIRED_GENEZIO_TYPES_VERSION_RANGE,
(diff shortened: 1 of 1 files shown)


src/commands/deploy.ts
```diff
@@ -119,27 +118,6 @@ export async function deployCommand(options: GenezioDeployOptions) {
...
-        if (
-            ENVIRONMENT === "dev" &&
-            configuration.backend &&
-            process.env["DISABLE_AB_TESTING"] !== "true"
-        ) {
-            const yamlConfig = await configIOController.read(/* fillDefaults= */ false);
-            if (!yamlConfig.backend) {
-                throw new UserError("No backend entry in genezio configuration file.");
-            }
-            yamlConfig.backend.cloudProvider = await performCloudProviderABTesting(
```
costinsin, line 133: Should we also remove `performCloudProviderABTesting` and tests associated to it (`tests/performAbTesting.test.ts`)?

costinsin (approved):

PR #1046, opened 2024-05-10
Revert default provider change back to aws

MagoDelBlocco:
Revert back the default provider change from `genezio-cloud` to `genezio-aws`.


andreia-oca (approved):
LGTM

PR #1059, opened 2024-05-13
auth for Go

stefandarius:
## Type of change

-   [x] 🍕 New feature

## Description

Added authentication capabilities for GO. Users can specify that a method should be authenticated using the `genezio: auth` comment/decorator above the method signature.

Each authenticated method should have `context.Context` as the first parameter. If the authentication succeeds, The `user` key of the context will be populated with a `*auth.User` structure, defined in [this library](https://github.com/Genez-io/auth).

A documentation update will be ready soon.

## Checklist


costinsin (approved):
LGTM

PR #1060, opened 2024-05-13
Fix Runtime handler URL composition

MagoDelBlocco:
-   [ ] 🐛 Bug Fix

The url returned by the runtime now has the functionId as part of the subdomain. Previously, it was the first element of the path. As a result, in order to take the method name, we need to take the first part of the path.


valighita:
Can you please edit the PR description and the commit message to explain the change?
I know what the PR is doing, but it should be clear from the title/description as well.

valighita (approved):
LGTM

PR #1061, opened 2024-05-14
remove cloudProvider from yaml

stefandarius:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description

This PR removes the cloudProvider property from `genezio.yaml`. Now, when deploying a project, the CLI will perform an additional request to the backend to get the appropriate cloud provider, meaning the current provider for already existing projects, and the default cloud provider (genezio cloud) for new projects.

Also removed AB testing functionality (which has become unused).

## Checklist


src/requests/getCloudProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const response: AxiosResponse<StatusOk<{ cloudProvider: CloudProviderIdentifier }>> =
+        await uninterceptedAxiosInstance({
+            method: "GET",
+            url: `${BACKEND_ENDPOINT}/projects/cloud-provider/${name}`,
+            headers: {
+                Authorization: `Bearer ${authToken}`,
+                "Accept-Version": `genezio-cli/${version}`,
+            },
+        });
```
costinsin, line 40: I think an error is automatically thrown when the code is not `2**` or `3**`. This will not silently fail, it will throw an unexpected error if the HTTP call is not successful.

costinsin (approved):
LGTM! Good job!

src/requests/getCloudProvider.ts
stefandarius: you are right. i will add a catch statement returning the default value

costinsin (approved):

vladiulianbogdan (approved):

andreia-oca (approved):
LGTM!

PR #1101, opened 2024-05-28
Remove node gyp dependencies

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
## Description
Because node gyp requires python and the C compiler to be installed, we had issues when users were trying to install genezio. The tree-sitter dependecy uses node-gyp. Tree sitter was used for parsing the source code of the user and checking for specific comments like `// genezio: deploy`. To fix this, i used babel for javascript/typescript files and I made a go parser for go files. The go parser can be found here https://github.com/Genez-io/go-parser 

## Checklist


src/utils/decorators/baseDecoratorExtractor.ts
```diff
@@ -2,16 +2,30 @@ import FileDetails from "../../models/fileDetails.js";
...
+                name: methodName,
+                decorators: [
+                    {
+                        name: "GenezioMethod",
+                        arguments: {
+                            type: "jsonrpc",
+                        },
+                    },
+                ],
+            };
```
stefandarius, line 28: I think this is a bit redundant because every method is exposed as jsonrpc if not explicitly marked otherwise, so we don't need to return an object here.

src/utils/decorators/baseDecoratorExtractor.ts
```diff
@@ -58,22 +72,25 @@ export abstract class DecoratorExtractor {
...
         let classType: string = "jsonrpc";
-        switch (genezioDeployArguments[0]) {
-            case "http": {
-                classType = "http";
-                break;
-            }
-            case "cron": {
-                classType = "cron";
-                break;
+        if (genezioDeployArguments.length > 0) {
```
stefandarius, line 81: this is also not needed, if the array is empty, the element at index 0 would be undefined, meaning it would not satisfy any cases of the switch statement

stefandarius (approved):

vladiulianbogdan (approved):

src/utils/decorators/baseDecoratorExtractor.ts
Virgil993: the problem is the genezio deploy arguments with this commentText would be undefined and it would throw and error arguments[0]

PR #1142, opened 2024-06-10
communicate directly with function processes and assure port consistency

stefandarius:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description

Functions need to be available at `/` instead of `/.functions/:functionName`. For that, we want to communicate directly with the processes spawned by genezio and ensure they will use the same port for each run of `genezio local`.

## Checklist


costinsin (approved):
Good job!

PR #1156, opened 2024-06-13
added cors headers for each response

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🐛 Bug Fix
-   [x] 🔥 Breaking change

## Description
In our runtime we would add cors headers automatically for all functions, this however poses a problem when we have express/faas functions that handle their own cors policies. To fix we stopped adding cors headers in the runtime and added cors headers in the runtimeHandler 

## Checklist

### DIFF
--- a/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
+++ b/src/bundlers/node/genezioRuntimeHandlerGenerator.ts
@@ -97,9 +97,9 @@ if (!genezioClass) {
             const http2CompliantHeaders = {};
             for (const header in event.headers) {
                 http2CompliantHeaders[header.toLowerCase()] = event.headers[header];
-                http2CompliantHeaders['Access-Control-Allow-Origin'] = '*';
-                http2CompliantHeaders['Access-Control-Allow-Headers'] = '*';
-                http2CompliantHeaders['Access-Control-Allow-Methods'] = '*';
+                http2CompliantHeaders['Access-Control-Allow-Origin'.toLowerCase()] = '*';
+                http2CompliantHeaders['Access-Control-Allow-Headers'.toLowerCase()] = '*';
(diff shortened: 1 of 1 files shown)


costinsin (approved):
LGTM

src/bundlers/node/genezioRuntimeHandlerGenerator.ts
```diff
@@ -34,7 +34,7 @@ if (!genezioClass) {
                 },
                 id: 0,
             }),
-            headers: { 'Content-Type': 'application/json', 'X-Powered-By': 'genezio' }
+            headers: { 'Content-Type': 'application/json', 'X-Powered-By': 'genezio','Access-Control-Allow-Origin': '*', 'Access-Control-Allow-Headers': '*', 'Access-Control-Allow-Methods': '*' }
```
valighita, line 37: Why are we adding the header in all the responses? Shouldn't it be present only when the method is OPTIONS?

Virgil993: I tried to put it only in OPTIONS but it doesn't work, I still get a CORS error, the only way for this to work is if the responses also have CORS headers present. I think it might have something to do with the fact that when we used the CloudFormationBuilder on Lambda we set the CORS policy for the ApiGateway like this



```
        this.addResource(apiGatewayResourceName, {
            Type: "AWS::ApiGatewayV2::Api",
            Properties: {
                Name: apiGatewayName,
                ProtocolType: "HTTP",
                Description: `API Gateway for Genezio Project ${apiGatewayName}}`,
                CorsConfiguration: {
                    AllowOrigins: ["*"],
                    AllowMethods: ["*"],
                    AllowHeaders: ["*"],
                    MaxAge: 10800,
                },
            },
        });
```

I i think the ApiGateway handled adding CORS to each lambda response.

We don't have the same mechanism for the Runtime so we need to put it manually in each response.

PR #1157, opened 2024-06-14
Upload user code on nextjs branch

radu1122:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Checklist

-   [ ] I have added tests;


src/genezio.ts
```diff
@@ -168,6 +168,7 @@ Use --backend to deploy only the backend application.`,
             });
             exit(1);
         });
+        process.exit(0);
```
costinsin, line 171: Why is this necessary? Isn't the function returning anyway and exiting the program?

src/commands/deploy/command.ts
```diff
@@ -4,21 +4,56 @@ import { debugLogger } from "../../utils/logging.js";
...
 import { nextJsDeploy } from "./nextjs.js";
+import path from "path";
+import { createTemporaryFolder, zipDirectory } from "../../utils/file.js";
+import { getPresignedURLForProjectCodePush } from "../../requests/getPresignedURLForProjectCodePush.js";
+import { uploadContentToS3 } from "../../requests/uploadContentToS3.js";
+import { YamlConfigurationIOController } from "../../yamlProjectConfiguration/v2.js";
 
 export async function deployCommand(options: GenezioDeployOptions) {
     await interruptLocalProcesses();
+    // create archive of the project
```
costinsin, line 15: This comment is not related to the code below it. I would remove it.

costinsin (approved):

src/genezio.ts
vladiulianbogdan: We have a problem on the build machine. If we launch `genezio deploy` from another process, it seems that the process does not exit and we suspect this is because of the stdin read. We do the `process.exit(0)` as a hack for now to unblock this. We should figure it out why the process does not exit.

costinsin (changes requested):
This PR introduces a bug that will not deploy the Next.js code if the user skips the CDN deployment spinner. This happens because if any key is pressed, `process.exit` will be called, not letting the code upload finish after deployment.

costinsin:
I have fixed the bug mentioned above in https://github.com/Genez-io/genezio/pull/1161.

costinsin (approved):

PR #1172, opened 2024-06-19
allow object body and query params in genezio local functions

stefandarius:
## Type of change

-   [x] 🐛 Bug Fix

## Description

Fixed a bug that broke genezio local if trying to return a body of type object from a function. Also, fixed query params which broke since we trimmed the URL for local functions.

## Checklist


costinsin (approved):
LGTM

PR #1192, opened 2024-06-24
Add information in the request about what stack a project is using

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

We need to be able to differentiate between projects using different technologies. `stack` is an array of strings that can contain all the technologies that a project is using. For now, we are able to differentiate only for Next.js.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/cloudAdapter/cloudAdapter.ts
+++ b/src/cloudAdapter/cloudAdapter.ts
@@ -62,6 +62,7 @@ export interface CloudAdapter {
         input: GenezioCloudInput[],
         projectConfiguration: ProjectConfiguration,
         cloudAdapterOptions: CloudAdapterOptions,
+        stack: string[],
     ): Promise<GenezioCloudOutput>;
     deployFrontend(
         projectName: string,
--- a/src/cloudAdapter/genezio/genezioAdapter.ts
+++ b/src/cloudAdapter/genezio/genezioAdapter.ts
@@ -115,6 +115,7 @@ export class GenezioCloudAdapter implements CloudAdapter {
         input: GenezioCloudInput[],
         projectConfiguration: ProjectConfiguration,
         cloudAdapterOptions: CloudAdapterOptions,
+        stack: string[] = [],
     ): Promise<GenezioCloudOutput> {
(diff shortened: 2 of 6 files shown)


src/cloudAdapter/cloudAdapter.ts
```diff
@@ -62,6 +62,7 @@ export interface CloudAdapter {
         input: GenezioCloudInput[],
         projectConfiguration: ProjectConfiguration,
         cloudAdapterOptions: CloudAdapterOptions,
+        stack: string[],
```
costinsin, line 65: Why is this a `string[]` instead of a `string`? How do you plan to use this feature in the future?

andreia-oca (approved):
LGTM

src/cloudAdapter/cloudAdapter.ts
andreia-oca: My understanding is that a project will have a list of technologies defined in `stack`.
E.G. A NextJs project with an Express Backend will have the `stack: [nextjs, express]`.

For now, this is only implemented for NextJs projects, hence the stack is only `[nextjs]`.

vladiulianbogdan: @andreia-oca is correct. That was the intention. Do you have any concerns, @costinsin ?

costinsin: Yes, I have. At the moment we either decide if the project is Next.js or classic deploy, so surely in at the moment there is no in between as @andreia-oca said. I am not sure how this will evolve in the future, but I am expecting to have one "stack" per project for now.

vladiulianbogdan: I am also not sure how things will evolve, but I can see it happening that we might need a project to have multiple technologies. That's why I've implemented like this, in case we need it, we won't have to change anything.

However, if we won't need it, it will still work.

costinsin (approved):

PR #1196, opened 2024-06-27
Add streaming support to Lambda faas handler

valighita:
## Type of change

-   [x] 🍕 New feature

## Description

Extend the Lambda handler to support response streaming.
Mock the `awslambda.streamifyResponse` function to use the stream created by the Genezio function runtime.

### DIFF
--- a/src/functionHandlerProvider/functionHandlerProvider.ts
+++ b/src/functionHandlerProvider/functionHandlerProvider.ts
@@ -2,4 +2,5 @@ import { FunctionConfiguration } from "../models/projectConfiguration.js";
 
 export interface FunctionHandlerProvider {
     getHandler(functionConfiguration: FunctionConfiguration): Promise<string>;
+    writeAdditionalFiles(path: string): Promise<void>;
 }
--- a/src/functionHandlerProvider/providers/AwsFunctionHandlerProvider.ts
+++ b/src/functionHandlerProvider/providers/AwsFunctionHandlerProvider.ts
@@ -1,10 +1,12 @@
+import { writeToFile } from "../../utils/file.js";
 import { FunctionConfiguration } from "../../models/projectConfiguration.js";
 import { FunctionHandlerProvider } from "../functionHandlerProvider.js";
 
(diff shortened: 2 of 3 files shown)


src/functionHandlerProvider/functionHandlerProvider.ts
```diff
@@ -2,4 +2,5 @@ import { FunctionConfiguration } from "../models/projectConfiguration.js";
 
 export interface FunctionHandlerProvider {
     getHandler(functionConfiguration: FunctionConfiguration): Promise<string>;
+    writeAdditionalFiles(path: string): Promise<void>;
```
costinsin, line 5: Does the naming of this function make sense for other Function Handler Providers like GCP or Azure? I feel like the naming of this can be more explicit.

costinsin (approved):
LGTM

src/functionHandlerProvider/functionHandlerProvider.ts
valighita: This is not particular to Lambda, it is just a way to generate additional files needed by the provider wrapper, if needed. If other providers don't need additional files, they don't need to do anything.

vladiulianbogdan: I think the whole interface should be rewritten. Instead of having two methods one that returns the `getHandler` and one that `writeAdditionalFiles`, I think we should have just one method that is called `writeHandler` and does everything.

src/functionHandlerProvider/providers/AwsFunctionHandlerProvider.ts
```diff
@@ -1,10 +1,24 @@
+import { writeToFile } from "../../utils/file.js";
 import { FunctionConfiguration } from "../../models/projectConfiguration.js";
 import { FunctionHandlerProvider } from "../functionHandlerProvider.js";
 
+const content = `global.awslambda = {
```
costinsin, line 5: Content is too generic, find a better naming.

costinsin (approved):

PR #1228, opened 2024-07-16
Prevent browser redirect at login if CI is set to true

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

If `CI` is set to true it means the genezio process runs in a continuous integration environment.
Trying to open the browser will make the process hang indefinitely.

This eats away GitHub actions minutes for failed logins because it will hang until the action timeouts or is cancelled.

## Checklist

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -15,6 +15,14 @@ export async function loginCommand(accessToken: string, logSuccesMessage = true)
         eventType: TelemetryEventTypes.GENEZIO_LOGIN,
     });
 
+    // If we are in a CI environment, we don't open the browser because it will hang indefinitely
+    if (process.env["CI"]) {
+        log.error(
+            "CI environment detected. Cannot open browser for authentication. Use `genezio login <token>` instead.",
+        );
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


src/commands/login.ts
```diff
@@ -55,6 +55,14 @@ export async function loginCommand(accessToken: string, logSuccesMessage = true)
...
             });
 
+            // If we are in a CI environment, we don't open the browser because it will hang indefinitely
+            if (process.env["CI"]) {
+                log.error(
+                    "CI environment detected. Cannot open browser for authentication. Use `genezio login <token>` instead.",
+                );
+                return;
+            }
+
```
costinsin, line 65: I would do this before opening the http server.

costinsin (approved):

src/commands/login.ts
andreia-oca: Nice suggestion, thanks :D

costinsin (changes requested):

PR #1284, opened 2024-08-14
[YAML] Inherit all the environment variables when running scripts

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

When executing `execa` with `env` set, all the exported env variables are overwritten.
We should also add `process.env` in the `env` field.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/utils/scripts.ts
+++ b/src/utils/scripts.ts
@@ -164,7 +164,14 @@ export async function runScript(
     }
 
     for (const script of scripts) {
-        await execaCommand(script, { cwd, shell: true, env: environment });
+        await execaCommand(script, {
+            cwd,
+            shell: true,
+            env: {
+                ...process.env,
+                ...environment,
+            },
+        });
     }
 }
 


src/utils/scripts.ts
```diff
@@ -164,7 +164,14 @@ export async function runScript(
     }
 
     for (const script of scripts) {
-        await execaCommand(script, { cwd, shell: true, env: environment });
+        await execaCommand(script, {
+            cwd,
+            shell: true,
+            env: {
+                ...process.env,
```
costinsin, line 171: I think this is automatically inherited https://github.com/sindresorhus/execa/blob/main/docs/api.md#optionsextendenv.

> If true, the subprocess uses both the [env](https://github.com/sindresorhus/execa/blob/main/docs/api.md#optionsenv) option and the current process' environment variables ([process.env](https://nodejs.org/api/process.html#processenv)). If false, only the env option is used, not process.env.

costinsin (approved):
LGTM

andreia-oca:
`process.env` is automatically inherited, this PR is not needed.

PR #1286, opened 2024-08-15
Throw error if frontend publish folder is missing

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Throw error if frontend `publish` folder is missing instead of just returning a warning.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -696,8 +696,9 @@ export async function deployFrontend(
         log.info(
             `Skipping frontend deployment for \`${frontend.path}\` because it has no publish folder in the YAML configuration. Check https://genezio.com/docs/project-structure/genezio-configuration-file for more details.`,
         );
-
-        return;
+        throw new UserError(
+            `Frontend named ${name} has no publish folder in the YAML configuration.`,
+        );
     }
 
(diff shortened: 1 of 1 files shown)


costinsin:
I think there are situations where you want to generate a SDK, but you don't want to publish anything. This is the case when you leave publish not defined.

costinsin:
Also this is not the right change if you intend to enforce publish. You need to change the `zod` definition.

radu1122 (approved):

vladiulianbogdan:
> I think there are situations where you want to generate a SDK, but you don't want to publish any frontend. This is the case when you leave publish not defined.

Could you use this in that scenario https://genezio.com/docs/cli-tool/cli-commands/generatesdk/? 

The thing is that it has some weird side effects the fact that you have a frontend without `publish`. For example if you would import [this](https://github.com/paulacionca/nodejs-serverless-function-express) project via the Github Integration, it will seem as if everything worked fine, but it is actually a misconfiguration.

costinsin (approved):
LGTM

PR #1324, opened 2024-08-23
Use the correct database type

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Enforce `AuthenticationDatabaseType` for setting a database for auth.

export enum AuthenticationDatabaseType {
    mongo = "mongodb",
    postgres = "postgresql",
}

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/utils.ts
+++ b/src/commands/deploy/utils.ts
@@ -11,7 +11,10 @@ import {
     SetAuthProvidersRequest,
     YourOwnAuthDatabaseConfig,
 } from "../../models/requests.js";
-import { DatabaseType } from "../../projectConfiguration/yaml/models.js";
+import {
+    AuthenticationDatabaseType,
+    DatabaseType,
+} from "../../projectConfiguration/yaml/models.js";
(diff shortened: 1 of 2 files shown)


src/commands/deploy/utils.ts
```diff
@@ -255,7 +258,7 @@ export async function enableAuthentication(
             {
                 enabled: true,
                 databaseUrl: databaseUri,
-                databaseType: authDatabase.type,
+                databaseType: authDatabase.type as AuthenticationDatabaseType,
```
costinsin, line 261: Why do you need the type cast? Can't we make the type system understand this type? Zod should attest that this type is really a auth db type.

costinsin (changes requested):

src/commands/deploy/utils.ts
andreia-oca: Yes, you are correct. I have updated the types to correctly enforce `AuthenticationDatabaseType` without any cast.
Thanks

PR #1339, opened 2024-08-27
When a frontend is deployed for the first time, attempt to set a redi…

andreia-oca:
…rect URL for the password

# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

I reverted in #1338 because the feature was not usable - for now I believe we should not add `settings.templates` to the yaml file because it will clutter the file unnecessary and it's not the best DX.

To support one-click deployment for templates that use `authentication/resetPassword/redirectUrl`, we are going to support a best effort approach where when deploying a frontend for the first time, we also set the `resetPassword/redirectUrl`.

The `redirectUrl` will not be updated if:
- a second subdomain is deployed
- a custom domain is set
- the authentication is enabled from UI
If any of these occur, the user should update/set the `redirecUrl` manually.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -72,6 +72,7 @@ import {
     uploadEnvVarsFromFile,
     processYamlEnvironmentVariables,
     enableAuthentication,
+    setDefaultPasswordRedirectUrl,
 } from "./utils.js";
 import { enableEmailIntegration } from "../../requests/integration.js";
 import { findAnEnvFile } from "../../utils/environmentVariables.js";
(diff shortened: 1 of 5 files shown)


src/commands/deploy/genezio.ts
```diff
@@ -751,6 +752,11 @@ export async function deployFrontend(
 
     frontend.subdomain = options.subdomain || frontend.subdomain;
 
+    // If authentication is enabled in the project - attempt to set a default password reset URL
+    if (configuration.services?.authentication) {
+        await setDefaultPasswordRedirectUrl(name, stage, frontend.subdomain || "");
+    }
```
costinsin, line 758: What if there is no backend deployed? This will fail because it will not find any project with this name. I would move this after creating the frontend so we could be sure that the environment exists.

costinsin (approved):

PR #1345, opened 2024-08-28
Add support for frontend v2 for nuxt

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Add CDN support for Nuxt.

## Checklist

-   [ ] I have added tests;


src/utils/file.ts
```diff
@@ -105,6 +105,7 @@ export async function getAllFilesFromCurrentPath(
 export async function zipDirectory(
     sourceDir: string,
     outPath: string,
+    includeHiddenFiles: boolean = false,
```
andreia-oca, line 108: I think this should be `true` by default.

So before this change `dot: true`, hence all of the previous method calls assume that hidden files will be included by default, right?

src/commands/deploy/utils.ts
```diff
@@ -716,3 +723,31 @@ export async function uploadEnvVarsFromFile(
...
+            "./node_modules",
+            "node_modules",
+            "node_modules/**",
+            "**/node_modules/**",
+        ],
+    );
+
+    await promiseZip;
+    const presignedUrlForProjectCode = await getPresignedURLForProjectCodePush(region, name, stage);
+    return uploadContentToS3(
```
andreia-oca, line 749: I think we should clean `tmpFolderProject` here with something like:
```
await deleteFolder(path.dirname(tmpFolderProject));
```

src/commands/deploy/utils.ts
```diff
@@ -716,3 +723,31 @@ export async function uploadEnvVarsFromFile(
...
+        [
+            "**/node_modules/*",
+            "./node_modules/*",
+            "node_modules/*",
+            "**/node_modules",
+            "./node_modules",
+            "node_modules",
+            "node_modules/**",
+            "**/node_modules/**",
+        ],
```
andreia-oca, line 744: It seems to me that the exclusion list is smaller than it was before refactoring.
I believe we are adding the `.git` directory and any other temporary directories used by vercel, opennnext in the zip.

Is this something we want?

I think we should at least keep ignoring the `.git` related stuff.

src/commands/deploy/genezio.ts
```diff
@@ -279,83 +278,7 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
...
-    const presignedUrlForProjectCode = await getPresignedURLForProjectCodePush(
-        configuration.region,
-        configuration.name,
-        options.stage,
-    );
-    await uploadContentToS3(
-        presignedUrlForProjectCode,
-        path.join(tmpFolderProject, "projectCode.zip"),
-    );
+    await uploadUserCode(configuration.name, configuration.region, options.stage);
```
costinsin, line 281: Thanks for doing this refactoring! It was DEFINITELY needed.

src/commands/deploy/utils.ts
costinsin: All temporary folders are already deleted at process exit:

https://github.com/Genez-io/genezio/blob/a5e3ea5878b5a41563738926dc30ab8e350ed837/src/index.ts#L40

src/commands/deploy/nuxt/deploy.ts
```diff
@@ -72,16 +97,121 @@ async function deployFunctions(config: YamlProjectConfiguration, stage?: string)
...
+                dot: true,
+                cwd: folder,
+            },
+            (err, files) => {
+                if (err) {
+                    reject(err);
+                    return;
+                }
+
+                const paths: CreateFrontendV2Path[] = files.map((file) => ({
```
costinsin, line 207: What if there are more than 250 files? 😥

costinsin (approved):
LGTM

src/commands/deploy/utils.ts
andreia-oca: My bad, disregard my comment

src/utils/file.ts
vladiulianbogdan: Doesn't matter. I changed in all places where this is used. The default value is for future usage and I think false is a more safe default value.

src/commands/deploy/utils.ts
vladiulianbogdan: All dot files and folders are ignored. That's why I did the change with `zipDirectory`.

src/utils/file.ts
andreia-oca: It's okay to let it be false, but can you also update the following references:

https://github.com/Genez-io/genezio/blob/dev/src/commands/bundle.ts#L117
https://github.com/Genez-io/genezio/blob/dev/src/bundlers/dart/dartBundler.ts#L214

andreia-oca (approved):
LGTM 🦭

src/commands/deploy/nuxt/deploy.ts
vladiulianbogdan: We cry. I'll add a task for this.

andreia-oca (approved):

PR #1348, opened 2024-08-30
Search ${{env.KEY}} in process.env 

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

This PR adds support for searching environment variables in `process.env` also.

This is available only for env vars declared in the `genezio.yaml` file with `${{env.MY_ENV}}`.

This is useful for testing commands or CI commands where the user runs:
```
MY_ENV=test genezio deploy
```
In our case is useful for running integration tests for the `services.authentication` feature.

Note: This pr contains a small refactoring for supporting a string prefix/suffix for genezio.yaml expressions (e.g. `https://${{<expression>}}/endpoint`)

## Checklist

-   [x] I have updated the documentation;
-   [x] I have added tests;

### DIFF
--- a/src/utils/strings.ts
+++ b/src/utils/strings.ts
@@ -35,3 +35,20 @@ export const cyan = "\x1b[36m" + reset;
 export const displayHint = function (message: string) {
     return `${colors.green(`Hint:`)} ${message}`;
 };
+
+/**
+ * Replace an expression of format <prefix>${{<variable>}}<suffix>
+ * to <prefix><value><suffix>
+ *
+ * ${{<variable>}} can be any alphanumeric string with special characters
+ * like /, -, ., _ and whitespace
+ *
+ * @param {string} resource - The original resource string
+ * @param {string} value - The value to replace the placeholder with.
+ * @returns {string} - The updated string with the placeholder replaced.
+ */
+export function replaceExpression(expression: string, value: string) {
+    // ${{<variable>}} can be any alphanumeric string with special characters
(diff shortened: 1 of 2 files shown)


costinsin (approved):

src/utils/strings.ts
```diff
@@ -35,3 +35,20 @@ export const cyan = "\x1b[36m" + reset;
...
+ * to <prefix><value><suffix>
+ *
+ * ${{<variable>}} can be any alphanumeric string with special characters
+ * like /, -, ., _ and whitespace
+ *
+ * @param {string} resource - The original resource string
+ * @param {string} value - The value to replace the placeholder with.
+ * @returns {string} - The updated string with the placeholder replaced.
+ */
+export function replaceExpression(expression: string, value: string) {
```
costinsin, line 50: What happens if there are multiple variables in the same expression string? You would only replace the first one, right?

andreia-oca: Yes, we don't support multiple expressions in the same string, at least for now seems there is no use case.

andreia-oca: For these cases

```
beginning${{<expression_1>}}something${{<expression_2>}}final
```

It will expand to
```
beginning<value>something${{<expression_2>}}final
```

PR #1397, opened 2024-09-24
Feat/support for containers

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Add support for deploying containers.

## Checklist

-   [ ] I have added tests;


src/constants.ts
costinsin, line 1: Drop this change

package.json
```diff
@@ -126,6 +126,7 @@
         "@types/cli-table": "^0.3.4",
         "@types/cors": "^2.8.17",
         "@types/decompress": "^4.2.7",
+        "@types/dockerode": "^3.3.31",
```
vladiulianbogdan, line 129: I need to remove this

src/commands/deploy/command.ts
```diff
@@ -83,5 +89,9 @@ function decideDeployType(): DeployType {
         }
     }
 
+    if (fs.existsSync("Dockerfile")) {
```
stefandarius, line 92: this means the CLI will assume a docker deployment whenever it detects a Dockerfile? if I would like to switch to classic deployment I should just delete de Dockerfile?

vladiulianbogdan: Yea, or move the Dockerfile or rename it... it is not ideal, but we wanted to get this started.

PR #1402, opened 2024-09-27
Sanitize frontend subdomain

stefandarius:
## Type of change

-   [x] 🧑‍💻 Improvement

## Description

This PR ensures that subdomains longer than 63 characters will never be generated

## Checklist

### DIFF
--- a/src/cloudAdapter/genezio/genezioAdapter.ts
+++ b/src/cloudAdapter/genezio/genezioAdapter.ts
@@ -27,7 +27,7 @@ import { calculateBiggestFiles } from "../../utils/calculateBiggestProjectFiles.
 import Table from "cli-table";
 import { UserError } from "../../errors.js";
 import { stdout } from "process";
-import { randomString } from "../../utils/strings.js";
+import { createHash } from "../../utils/strings.js";
 
 const BUNDLE_SIZE_LIMIT = 256901120;
 async function handleBigElementSizeError(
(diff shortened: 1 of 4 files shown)


src/cloudAdapter/genezio/genezioAdapter.ts
```diff
@@ -207,7 +208,11 @@ export class GenezioCloudAdapter implements CloudAdapter {
     ): Promise<string> {
         const finalStageName =
             stage != "" && stage != "prod" ? `-${stage.replaceAll(/[/_.]/gm, "-")}` : "";
-        const finalSubdomain = frontend.subdomain + finalStageName;
+        let finalSubdomain = frontend.subdomain + finalStageName;
+        if (finalSubdomain.length > 63) {
+            debugLogger.debug("Subdomain is too long. Generating random subdomain.");
+            finalSubdomain = frontend.subdomain?.substring(0, 56) + "-" + randomString(6);
```
vladiulianbogdan, line 214: If I deploy the same frontend a second time, will I get a new URL?

stefandarius: auch, you are right. that's an error on my side. I'll have to think of a way to ensure the same suffix is generated. maybe a hash of the stage name?

vladiulianbogdan (approved):

src/cloudAdapter/genezio/genezioAdapter.ts
vladiulianbogdan: Yea, a hash sounds like a good idea.

vladiulianbogdan (approved):

PR #1415, opened 2024-10-08
[FAAS]: Add support python functions

cristim67:
## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Add support python functions

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -619,7 +619,6 @@ export async function functionToCloudInput(
     const unzippedBundleSize = await getBundleFolderSizeLimit(tmpFolderPath);
 
     // add the handler to the temporary folder
-    // check if there already is an index.mjs file in user's code
     let entryFileName = entryFileFunctionMap[functionElement.language as Language];
     while (fs.existsSync(path.join(tmpFolderPath, entryFileName))) {
         debugLogger.debug(
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -602,8 +602,8 @@ export async function functionToCloudInput(
     );
 
     // copy everything to the temporary folder
+    await fsExtra.copy(path.join(backendPath, functionElement.path), tmpFolderPath);
     if (functionElement.language === "js" || functionElement.language === "ts") {
-        await fsExtra.copy(path.join(backendPath, functionElement.path), tmpFolderPath);
         if (fsExtra.pathExistsSync(path.join(tmpFolderPath, "node_modules", ".pnpm"))) {
             await fsExtra.remove(path.join(tmpFolderPath, "node_modules", ".pnpm"));
             await fsExtra.copy(
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -489,7 +489,7 @@ export async function deployClasses(
             functionToCloudInput(
                 f,
                 backend.path,
-                undefined,
+                /* outputDir */ undefined,
                 configuration.backend?.language.packageManager,
             ),
     );
--- a/src/commands/deploy/genezio.ts
+++ b/src/commands/deploy/genezio.ts
@@ -52,6 +52,7 @@ import {
 import { YamlConfigurationIOController } from "../../projectConfiguration/yaml/v2.js";
 import {
     AuthenticationEmailTemplateType,
+    entryFileFunctionMap,
     FunctionType,
     Language,
 } from "../../projectConfiguration/yaml/models.js";
@@ -63,7 +64,10 @@ import { writeSdk } from "../../generateSdk/sdkWriter/sdkWriter.js";
 import { reportSuccessForSdk } from "../../generateSdk/sdkSuccessReport.js";
 import { isLoggedIn } from "../../utils/accounts.js";
 import { loginCommand } from "../login.js";
-import { AwsFunctionHandlerProvider } from "../../functionHandlerProvider/providers/AwsFunctionHandlerProvider.js";
+import {
+    AwsFunctionHandlerProvider,
+    AwsPythonFunctionHandlerProvider,
+} from "../../functionHandlerProvider/providers/AwsFunctionHandlerProvider.js";
 import fsExtra from "fs-extra/esm";
 import { getLinkedFrontendsForProject } from "../../utils/linkDatabase.js";
 import { getCloudProvider } from "../../requests/getCloudProvider.js";
@@ -575,12 +579,17 @@ export async function functionToCloudInput(
     backendPath: string,
     outputDir?: string,
 ): Promise<GenezioCloudInput> {
-    if (functionElement.language !== "js" && functionElement.language !== "ts") {
+    const supportedFunctionLanguages = ["js", "ts", "python"];
+
+    if (!supportedFunctionLanguages.includes(functionElement.language)) {
         throw new UserError(
-            `The language ${functionElement.language} is not supported for functions. Only JavaScript and TypeScript are supported.`,
+            `The language ${functionElement.language} is not supported for functions. Supported languages are: ${supportedFunctionLanguages.join(", ")}`,
         );
     }
-    const handlerProvider = getFunctionHandlerProvider(functionElement.type);
+    const handlerProvider = getFunctionHandlerProvider(
+        functionElement.type,
+        functionElement.language as Language,
+    );
 
     if (outputDir && !fs.existsSync(outputDir)) {
         debugLogger.debug(`Creating output directory ${outputDir}`);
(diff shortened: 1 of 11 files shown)


src/commands/deploy/genezio.ts
```diff
@@ -207,6 +211,25 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
             throw error;
         });
 
+        // Handle Python requirements.txt file
+        if (configuration.backend.language.name === Language.python) {
+            const requirementsPath = path.join(backendCwd, "requirements.txt");
+
+            if (fs.existsSync(requirementsPath)) {
```
stefandarius, line 218: i don't really like this logic being here, the `genezioDeploy` function should not have any language specific logic. is there a way you could include this step in the python bundler?

stefandarius: also, don't think we should modify the project structure of our users code. if you want to create a folder containing requirements, do this in the temporary folder created in the next steps (I will leave a comment there also)

src/models/projectOptions.ts
```diff
@@ -12,5 +14,11 @@ export type NodeOptions = {
...
 };
 
+export type PythonOptions = {
+    pythonRuntime: PythonRuntime;
+    architecture: Architecture;
+};
+
 export const supportedNodeRuntimes = ["nodejs20.x"] as const;
 export const supportedArchitectures = ["arm64", "x86_64"] as const;
+export const supportedPythonRuntimes = ["python3.13.x"] as const;
```
andreia-oca, line 24: We should more flexible with python versions because python releases more often than node (almost yearly) and also I believe users don't update often (I have python 3.10 installed on my machine).

As a rule of thumb we should follow the Python releases cycle [1]. I would be happy to as flexible as possible for our users and don't force them to upgrade their python - so I would support >= 3.9.

[1] https://devguide.python.org/versions/

src/commands/deploy/genezio.ts
```diff
@@ -594,27 +622,30 @@ export async function functionToCloudInput(
...
-            {
-                dereference: true,
-            },
-        );
+
+    // Handle JS/TS functions with pnpm
+    if (functionElement.language === "js" || functionElement.language === "ts") {
+        if (fsExtra.pathExistsSync(path.join(tmpFolderPath, "node_modules", ".pnpm"))) {
+            await fsExtra.remove(path.join(tmpFolderPath, "node_modules", ".pnpm"));
+            await fsExtra.copy(
```
stefandarius, line 630: this is what i was talking about. we should do the same with the python packages

src/commands/deploy/genezio.ts
```diff
@@ -817,13 +848,32 @@ export function getCloudAdapter(provider: CloudProviderIdentifier): CloudAdapter
     }
 }
 
-export function getFunctionHandlerProvider(functionType: FunctionType): AwsFunctionHandlerProvider {
+export function getFunctionHandlerProvider(
```
stefandarius, line 851: i think @radu1122 added a new provider/function type: `httpServer`. we should make sure to update this function accordingly

stefandarius (approved):

src/packageManagers/pip.ts
```diff
@@ -0,0 +1,56 @@
...
+        $({ cwd }).sync`pip install ${packages.join(" ")}`;
+    }
+
+    async link(packages: string[] = [], cwd?: string) {
+        await $({ cwd })`pip install -e ${packages.join(" ")}`;
+    }
+
+    async publish(cwd?: string) {
+        // For Python, publishing usually involves using `twine` to upload to PyPI
+        await $({ cwd })`python setup.py sdist bdist_wheel`;
```
andreia-oca, line 22: This might fail because users can run `python` or `python3` - on my machine I don't have the `python` command at all, I have only `python3`.

Historically, this is because of machines that had installed both python2 and python3. Usually `python` is an alias to one of them, but there are systems that don't have `python` at all.

I don't think it's necessary to fix this now, but you should add a TODO comment to describe how this might fail.

src/projectConfiguration/yaml/models.ts
```diff
@@ -35,3 +35,13 @@ export enum FunctionType {
...
+
+export const entryFileFunctionMap = {
+    js: "index.mjs",
+    ts: "index.mjs",
+    swift: "index.swift",
+    python: "index.py",
+    dart: "index.dart",
+    kt: "index.kt",
+    go: "index.go",
+};
```
andreia-oca, line 47:
```suggestion
export const entryFileFunctionMap = {
    js: "index.mjs",
    ts: "index.mjs",
    python: "index.py",
    go: "main.go",
};
```

andreia-oca: It's a bit confusing to have entry files for languages we are not supporting in production now.

src/commands/deploy/genezio.ts
```diff
@@ -207,6 +211,25 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
...
+                    const pythonPackagesPath = path.join(backendCwd, "packages");
+                    const pipInstallCommand = `pip install -r requirements.txt -t ${pythonPackagesPath}`;
+                    await runScript(pipInstallCommand, backendCwd);
+                } else {
+                    debugLogger.warn(
+                        "The requirements.txt file is empty. Skipping the pip install command.",
+                    );
+                }
+            }
+        }
```
andreia-oca, line 231: This can fail in soo many ways :)) virtualenv not configured, clashing with other global python dependencies, not using pip at all as a package manager (some devs are using `poetry`).

For now I would rely on the `genezio.yaml` file to add a script to install deps before deploying as it is for node.
```
backend:
  scripts:
    deploy: pip install -r requirements
```

Handling automating installs for dependencies will for sure be a task in the near future, but I think that now just running `pip instal -r ...` can potentially block some projects.

andreia-oca (approved):
Please add an integration test for this also

src/commands/deploy/genezio.ts
cristim67: Unlike in Node.js, where it's common to install dependencies locally to a project, in Python, it's more common to install packages globally. However, virtual environments are still a widely used option for project-specific dependencies. Given the many potential edge cases (global vs. virtual environment), the safest solution would be to focus on the dependencies listed in the requirements.txt file and install them ourselves. Some IDEs even automatically update the requirements.txt, making this the most reliable approach. 

I add poetry package manager also.

src/commands/deploy/genezio.ts
```diff
@@ -481,7 +485,13 @@ export async function deployClasses(
     );
 
     const functionsResultArray: Promise<GenezioCloudInput>[] = projectConfiguration.functions.map(
-        (f) => functionToCloudInput(f, backend.path),
+        (f) =>
+            functionToCloudInput(
+                f,
+                backend.path,
+                undefined,
```
andreia-oca, line 492:
```suggestion
                /* outputDir */ undefined,
```

andreia-oca: For readability, otherwise is hard to see what argument is undefined.

src/projectConfiguration/yaml/v2.ts
```diff
@@ -87,28 +89,27 @@ function parseGenezioConfig(config: unknown) {
...
+            );
+        }, "The handler should be in the format 'file.extension'. example: index.js / index.mjs / index.cjs / index.py"),
+        type: zod.nativeEnum(FunctionType).default(FunctionType.aws),
+    })
+    .refine(
             ({ type, handler }) => !(type === FunctionType.aws && !handler),
             "The handler is mandatory for type aws functions.",
         );
-
+
```
andreia-oca, line 112:
```suggestion
```

andreia-oca: Remove trailing white space please

src/commands/deploy/genezio.ts
```diff
@@ -818,15 +856,34 @@ export function getCloudAdapter(provider: CloudProviderIdentifier): CloudAdapter
...
+            };
+
+            const key = `${functionType}-${language}`;
+            const provider = providerMap[key];
+
+            if (provider) {
+                return provider;
+            } else {
+                throw new UserError(
+                    `Unsupported language: ${language} for AWS function. Supported languages are: python, js, ts.`,
```
andreia-oca, line 878:
```suggestion
                    `Unsupported language: ${language} for AWS-compatible function. Supported languages are: python, js, ts.`,
```

src/commands/deploy/genezio.ts
```diff
@@ -818,15 +856,34 @@ export function getCloudAdapter(provider: CloudProviderIdentifier): CloudAdapter
...
+                );
+            }
+        }
         case FunctionType.httpServer:
             return new HttpServerHandlerProvider();
         default:
             throw new UserError(
-                `Unsupported function type: ${functionType}. Supported providers are: aws`,
+                `Unsupported function type: ${functionType}. Supported providers are: aws.`,
```
andreia-oca, line 886:
```suggestion
                `Unsupported function type: ${functionType}. Supported providers are: aws, httpServer.`,
```

andreia-oca (approved):

PR #1416, opened 2024-10-08
Deploy docker only when an option is passed

vladiulianbogdan:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

To avoid any backwards compatibility issue, we deploy container only when an option is passed.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/genezio.ts
+++ b/src/genezio.ts
@@ -149,6 +149,11 @@ program
     .option("--env <envFile>", "Load environment variables from a given file", undefined)
     .option("--stage <stage>", "Set the environment name to deploy to", "prod")
     .option(
+        "--image <image>",
+        "Path to the container image file. If this option is set, it indicates a container-based deployment.",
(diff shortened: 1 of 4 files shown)


src/genezio.ts
```diff
@@ -148,6 +148,11 @@ program
     )
     .option("--env <envFile>", "Load environment variables from a given file", undefined)
     .option("--stage <stage>", "Set the environment name to deploy to", "prod")
+    .option(
+        "--image <image>",
+        "Path to the container image file. If this option is set, it indicates a container-based deployment.",
```
valighita, line 153: It should be to the container Dockerfile, or build file, not image file.

PR #1435, opened 2024-10-15
Add support for genezio local -> python

cristim67:
## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Add support for genezio local for FAAS Python.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/bundlers/utils.ts
+++ b/src/bundlers/utils.ts
@@ -87,3 +87,80 @@ export async function bundle(
     debugLogger.debug(`The bundling process finished successfully for file ${element.path}.`);
     return output;
 }
+
+export function getLocalFunctionWrapperCode(handler: string, entry: string, language: string) {
+    switch (language) {
+        case "js":
+        case "ts":
+            return `import { ${handler} as userHandler } from "./${entry}";
+
+import http from "http";
+
+const port = process.argv[2];
+
+const server = http.createServer((req, res) => {
+  if (req.method === 'POST') {
+    let body = '';
+    req.on('data', chunk => {
+      body += chunk.toString();
+    });
+    req.on('end', () => {
+      res.writeHead(200, {'Content-Type': 'text/plain'});
+      const jsonParsedBody = JSON.parse(body);
+
+      userHandler(jsonParsedBody).then((response) => {
+        res.end(JSON.stringify(response));
+    })
+    });
+  } else {
+    res.writeHead(404, {'Content-Type': 'text/plain'});
+    res.end('404 Not Found');
+  }
+});
+
+server.listen(port, () => {
+});
+`;
+        case "python":
+            return `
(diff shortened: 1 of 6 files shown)


src/bundlers/utils.ts
```diff
@@ -87,3 +87,80 @@ export async function bundle(
     debugLogger.debug(`The bundling process finished successfully for file ${element.path}.`);
     return output;
 }
+
+export function getLocalFunctionWrapperCode(handler: string, entry: string, language: string) {
```
andreia-oca, line 91: [nit] We can be more strict with the `language` here:

```suggestion
export function getLocalFunctionWrapperCode(handler: string, entry: string, language: Language) {
```

src/bundlers/utils.ts
```diff
@@ -87,3 +87,80 @@ export async function bundle(
...
+    switch (language) {
+        case "js":
+        case "ts":
```
andreia-oca, line 94:
```suggestion
        case "Language.js":
        case "Language.ts":
```

src/bundlers/utils.ts
```diff
@@ -87,3 +87,80 @@ export async function bundle(
...
+        httpd.serve_forever()
+    except KeyboardInterrupt:
+        sys.stdout.flush()  
+        httpd.server_close()
+
+if __name__ == "__main__":
+    run()
+`;
+        default:
+            return ``;
```
andreia-oca, line 164: This default might hide a subtle error because it's silent - we should never reach this in silence :))

On default I would throw an error:
```
throw new Error(`${language} is not supported yet for testing locally`);
```

src/commands/local.ts
```diff
@@ -629,16 +632,28 @@ async function startProcesses(
 
             await writeToFile(
                 path.join(tmpFolder),
-                "local_function_wrapper.mjs",
-                getLocalFunctionWrapperCode(functionInfo.handler, functionInfo.entry),
+                `local_function_wrapper.${entryFileFunctionMap[functionInfo.language as keyof typeof entryFileFunctionMap].split(".")[1]}`,
```
andreia-oca, line 635: This is kinda ugly :)) For readability and because you use it in multiple places, I would create a helper function that creates filenames like `local_function_wrapper.mjs`, `local_function_wrapper,py` or `index-random.mjs`:

```
function getFunctionEntryFilename(language: Language, filename: string): string {
    const extension = entryFileFunctionMap[language as keyof typeof entryFileFunctionMap].split(".")[1];
    return `${filename}.${extension}`;
}
```

andreia-oca (approved):
Looks good!

PR #1456, opened 2024-10-22
Add python httpServer handler

cristim67:
## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Checklist

### DIFF
--- a/src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
+++ b/src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
@@ -110,53 +110,66 @@ export class HttpServerPythonHandlerProvider implements FunctionHandlerProvider 
         functionConfiguration: FunctionConfiguration,
     ): Promise<void> {
         const handlerContent = `
-import os
-from importlib import import_module
-from http.server import BaseHTTPRequestHandler, HTTPServer
-import threading
-import asyncio
+from io import BytesIO
+import traceback
+from ${functionConfiguration.entry.split(".")[0]} import app as application
 
-# Import the original app.py, but it will automatically start the server
-app = import_module("${functionConfiguration.entry.split(".")[0]}")
-
-server = None
-original_create_server = HTTPServer
-
-# Override the HTTP server creation to store the reference to the server
-def create_server(*args, **kwargs):
-    global server
-    server = original_create_server(*args, **kwargs)
-    return server
-
-class CustomHTTPRequestHandler(BaseHTTPRequestHandler):
-    def handle_request(self, event):
-        try:
-            self.command = event['http']['method']
-            self.path = f"{event['http']['path']}{event['url'].get('search', '')}"
-            self.headers = event.get('headers', {})
-            self.rfile = event['body'].encode() if isinstance(event['body'], str) else event['body']
-            self.client_address = (event['http']['sourceIp'], 0)
-            
-            # Emit the request to the server
-            server.process_request(self, None)
-        
-        except Exception as e:
-            self.send_response(500)
-            if os.getenv('GENEZIO_DEBUG_MODE') == 'true':
-                self.end_headers()
-                self.wfile.write(str(e).encode())
-            else:
-                self.end_headers()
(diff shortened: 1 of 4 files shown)


src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
```diff
@@ -102,3 +102,118 @@ export { handler };`;
...
+    }
+`;
+
+        await writeToFile(outputPath, handlerFileName, handlerContent);
+    }
+
+    // NOT USED
+    async getLocalFunctionWrapperCode(handler: string, entry: string): Promise<string> {
+        return `${handler} ${entry}`;
+    }
```
andreia-oca, line 218: If `not used`, why is it here? :))

If it should be here because it's called for these object types - at least it should be obvious it's not used - return `undefined` or maybe, `false`.

Otherwise - someday, it will be used, regardless the comment :))

src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
```diff
@@ -102,3 +102,118 @@ export { handler };`;
...
+    wsgi_headers = {}
+    for header, value in headers.items():
+        wsgi_headers[f"HTTP_{header.upper().replace('-', '_')}"] = value
+    return wsgi_headers
+
+def format_response(response_body, headers, status_code):
+    """
+    Formats the response to return, ensuring the content type is handled appropriately.
+    """
+    content_type = headers.get('Content-Type', 'text/html')
```
andreia-oca, line 191: What happens for `application/json` bodies?

src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
```diff
@@ -102,3 +102,118 @@ export { handler };`;
...
+            "headers": {"Content-Type": "text/plain"},
+            "error": str(e)
+        }
+
+def create_wsgi_environ(event, http_info, path):
+    """
+    Creates a WSGI environment based on the HTTP information received from the event.
+    """
+    return {
+        'REQUEST_METHOD': http_info.get('method', 'GET'),
```
andreia-oca, line 165: Hmm, this should be hardcoded to `GET`?
What happens if the function should handle post, delete etc?

cristim67: Here `GET` is a default value if the `http.method` attribute is not found in the `HTTP` event.

As long as `http.method` is specified, the assignment is dynamically. In flask for example, the method is specified and if it's not, the method will be `GET`:

```
@app.route('/')
def home():
    return 'Hello, World!'
    
@app.route('/post-data', methods=['POST'])
def post_data():
    data = request.get_json()
    if not data:
        return {'error': 'No data found in request'}, 400
    print(f"Received data: {data}")
    return data
```

src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
cristim67: It will be decoded as `utf-8`

cristim67: Also, I've added an example route for this: 

`@app.route('/post-data', methods=['POST'])
def post_data():
    data = request.get_json()
    if not data:
        return {'error': 'No data found in request'}, 400
    print(f"Received data: {data}")
    return data`

https://github.com/Genez-io/flask-getting-started

src/functionHandlerProvider/providers/HttpServerHandlerProvider.ts
cristim67: I don't necessarily have a context, I copied it from node (HttpServerHandlerProvider)

I think definition is needed because it is derived from an interface, but yes, I would return undefined. Is it ok for nodejs too @radu1122 ?

andreia-oca (approved):

PR #1495, opened 2024-11-08
Add cron service

Virgil993:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature

## Description

This PR adds the cron service to the yaml file and handles the local and deploy flows for crons that use genezio functions

## Checklist


src/commands/deploy/genezio.ts
```diff
@@ -230,6 +234,55 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
...
+                    });
+                    cronFound = true;
+                    break;
+                }
+            }
+            if (!cronFound) {
+                throw new UserError(
+                    `Function ${yamlFunctionName} not found in the deployed functions. Please make sure the function is deployed before adding it to the cron.`,
+                );
+            }
```
andreia-oca, line 267: This is too much logic here to prevent registering a cron without a function.

It would be better to attempt to register the cron and let the backend fail if there is no function deployed.

This will simplify a lot this code and improve performance (no need to fetch all the functions, iterate through them etc.)

Please rewrite this section with this philosophy in mind.

src/commands/deploy/genezio.ts
```diff
@@ -230,6 +234,55 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
...
+            throw new UserError("Something went wrong while syncing the cron jobs.");
+        });
+    } else {
+        await syncCrons({
+            projectName: projectName,
+            stageName: options.stage || "prod",
+            crons: [],
+        }).catch(() => {
+            throw new UserError("Something went wrong while syncing the cron jobs.");
+        });
```
andreia-oca, line 283: I am bit confused here - what should `syncCrons` do and why is it called twice?

andreia-oca: And also the errors messages should be improved - `Something went wrong` is not helping the user to find a solution - what he/she should do? Try to redeploy, try to debug, file an issue on github, contact support?

src/commands/local.ts
```diff
@@ -1051,6 +1055,49 @@ function startCronJobs(
         }
     }
 
+    if (yamlProjectConfiguration.services && yamlProjectConfiguration.services.crons) {
+        for (const cronService of yamlProjectConfiguration.services.crons) {
+            const functionName = await evaluateResource(
+                yamlProjectConfiguration,
+                cronService.function,
+                "local",
```
vladiulianbogdan, line 1063: what is this "local"?

src/commands/local.ts
```diff
@@ -1051,6 +1055,49 @@ function startCronJobs(
...
+                undefined,
+                {
+                    isLocal: true,
+                    port: port,
+                },
+            );
+            const endpoint = cronService.endpoint?.replace(/^\//, "");
+            const cronString = cronService.schedule;
+            const baseURL = `http://localhost:${port}/.functions/function-${functionName}`;
```
vladiulianbogdan, line 1072: I think this is prone to error in the future. If we change the `genezio local` command implementation and we serve the functions from a different link, this is very likely to be broken. The developer who will do that is likely to forget to change also here.

vladiulianbogdan (approved):

src/commands/deploy/genezio.ts
vladiulianbogdan: Probably it's called twice to remove the existing crons, if there are any.

src/utils/yaml.ts
```diff
@@ -10,3 +11,16 @@ export function generateRandomSubdomain(): string {
...
+export function isUnique(units: YamlCron[] | YamlFunction[], fieldName: FieldName): boolean {
+    const unitMap = new Map<string, string>();
+    for (const unit of units || []) {
+        if (unitMap.has(unit[fieldName])) {
+            return false;
+        }
+        unitMap.set(unit[fieldName], unit[fieldName]);
+    }
+    return true;
+}
```
andreia-oca, line 26: This method should be more general and flexible - ideally we would want to call it:
```
isUnique(anyZodArray, anyFieldName);
```

To achieve this you can change the following:
- use generic types instead of FieldName, YamlCron, YamlFunction. This should do the trick:
```
export function isUnique<T>(array: T[] | undefined, fieldName: keyof T): boolean 
```

- I think a Set instead of a Map makes more sense in this case (it's a tad more optimal because we will add a single value and the Set will automatically reject duplicate values)
```
export function isUnique<T>(units: T[] | undefined, field: keyof T): boolean {
    const unitSet = new Set<T[keyof T]>();
    ...
   unitSet.add(unit[fieldName]);
   ...
}
```

andreia-oca (approved):
Great job - this looks much more clean and readable now.

I have a few small improvements requests here and there, other than that it's good to be merged 🎉

PR #1498, opened 2024-11-08
Improve getting the entryfile for backends

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

Instead of assuming `packageJson.main` or `index.mjs` is the entryfile name, search for it in a grep recursive style.

When the first suitable candidate is found, we stop the search.
We exclude uninteresting directories such as `.git`, `node_modules`.

## Checklist

-   [x] I have added tests;

### DIFF
--- a/src/commands/analyze/command.ts
+++ b/src/commands/analyze/command.ts
@@ -11,11 +11,11 @@ import {
     isReactComponent,
     isViteComponent,
     isServerlessHttpBackend,
-    getEntryfile,
     isVueComponent,
     isAngularComponent,
     isSvelteComponent,
     isContainerComponent,
+    findEntryFile,
 } from "./frameworks.js";
 import { readOrAskConfig } from "../deploy/utils.js";
 import { getPackageManager, PackageManagerType } from "../../packageManagers/packageManager.js";
@@ -32,6 +32,7 @@ import {
 import { FunctionType, Language } from "../../projectConfiguration/yaml/models.js";
 import { report } from "./outputUtils.js";
 import { isCI } from "../../utils/process.js";
+import { EXPRESS_PATTERN, FASTIFY_PATTERN, SERVERLESS_HTTP_PATTERN } from "./constants.js";
 
 // backend javascript: aws-compatible functions, serverless-http functions, express, fastify
 // backend typescript: aws-compatible functions, serverless-http functions, express, fastify
@@ -64,6 +65,8 @@ export const DEFAULT_CI_FORMAT = SUPPORTED_FORMATS.JSON;
 
 export const KEY_FILES = ["package.json", "Dockerfile"];
 export const EXCLUDED_DIRECTORIES = ["node_modules", ".git", "dist", "build"];
+export const NODE_DEFAULT_ENTRY_FILE = "index.mjs";
+export const EXTENSIONS = [".js", ".ts", ".mjs", ".jsx", ".tsx"];
 
 // The analyze command has 2 side effects:
 // 1. It creates a new yaml with the detected components
@@ -104,7 +107,14 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
         };
 
         if (await isServerlessHttpBackend(contents)) {
-            const entryfile = await getEntryfile(contents);
+            const entryFile = await findEntryFile(
+                componentPath,
+                contents,
+                SERVERLESS_HTTP_PATTERN,
+                NODE_DEFAULT_ENTRY_FILE,
+            );
+            debugLogger.debug("Serverless HTTP entry file found:", entryFile);
+
             // TODO: Add support for detecting and building typescript backends
             // const isTypescriptFlag = await isTypescript(contents);
 
@@ -124,7 +134,7 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
                         path: ".",
                         // TODO: This is hardcoded because there are great chances that this indeed called `handler`
                         handler: "handler",
-                        entry: entryfile,
+                        entry: entryFile,
                         type: FunctionType.aws,
                     },
                 ],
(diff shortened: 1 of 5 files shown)


src/commands/analyze/command.ts
```diff
@@ -65,6 +66,8 @@ export const DEFAULT_CI_FORMAT = SUPPORTED_FORMATS.JSON;
 
 export const KEY_FILES = ["package.json", "Dockerfile"];
 export const EXCLUDED_DIRECTORIES = ["node_modules", ".git", "dist", "build"];
+export const NODE_DEFAULT_ENTRY_FILE = "index.mjs";
+export const EXTENSIONS = [".js", ".ts", ".mjs", ".mjs", ".jsx", ".tsx"];
```
cristim67, line 70: in `EXTENSIONS` you have a duplicate value `.mjs`

src/commands/analyze/frameworks.ts
```diff
@@ -35,15 +39,95 @@ export async function isTypescript(contents: Record<string, string>): Promise<bo
...
+        if (entry.isDirectory()) {
+            // Skip excluded directories
+            if (EXCLUDED_DIRECTORIES.includes(entry.name)) continue;
+
+            // Recursively search within subdirectories
+            const result = await findFileByPatterns(fullPath, patterns, extensions);
+            if (result) return result;
+        } else if (entry.isFile() && extensions.some((ext) => entry.name.endsWith(ext))) {
+            // Check if the file content matches all given patterns
+            const content = await fs.readFile(fullPath, "utf-8");
```
cristim67, line 88: Files are read twice: once during the initial population of `contents`, and again when pattern matching. Instead, we could leverage `contents` to store file contents on the first read and reuse them for pattern matching

```
        if (!contents[entry.name]) {
            contents[entry.name] = await fs.readFile(fullPath, "utf-8");
        }
        const content = contents[entry.name];
```

src/commands/analyze/frameworks.ts
```diff
@@ -35,15 +39,95 @@ export async function isTypescript(contents: Record<string, string>): Promise<bo
...
+): Promise<string | undefined> {
+    const entries = await fs.readdir(directory, { withFileTypes: true });
+
+    const contents: Record<string, string> = {};
+
+    // Populate contents with package.json and any files needed for direct access
+    for (const entry of entries) {
+        const fullPath = path.join(directory, entry.name);
+        if (entry.isFile()) {
+            const content = await fs.readFile(fullPath, "utf-8");
```
cristim67, line 71: You could check the file extension before reading contents to skip unnecessary reads, reducing processing time

cristim67 (approved):
LGTM

src/commands/analyze/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// We should take into account:
+//  * common methods called to instantiate servers/conections/components such as app.listen()
+//  * common imports such as import Fastify from 'fastify' or require('fastify') (ESM and CJS)
+
+// For each file there is a list of patterns all conditions should be matched to consider the file as a candidate
+
+export const FASTIFY_PATTERN = [
+    /app\.listen|fastify\.listen/,
+    /import\s+Fastify\s+from\s+['"]fastify['"]|require\(['"]fastify['"]\)/,
```
costinsin, line 11: why capital f?

```suggestion
    /import\s+fastify\s+from\s+['"]fastify['"]|require\(['"]fastify['"]\)/,
```

src/commands/analyze/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+];
+
+export const EXPRESS_PATTERN = [
+    /app\.listen/,
+    /import\s+express\s+from\s+['"]express['"]|require\(['"]express['"]\)/,
+];
+
+export const SERVERLESS_HTTP_PATTERN = [
+    /import\s+Serverless\s+from\s+['"]serverless-http['"]|require\(['"]serverless-http['"]\)/,
```
costinsin, line 20: why capital s?

```suggestion
    /import\s+serverless\s+from\s+['"]serverless-http['"]|require\(['"]serverless-http['"]\)/,
```

src/commands/analyze/frameworks.ts
```diff
@@ -35,15 +39,95 @@ export async function isTypescript(contents: Record<string, string>): Promise<bo
...
     if (!contents["package.json"]) {
-        return "index.mjs";
+        return undefined;
     }
 
-    // TODO Improve this - the entry file might not be defined in the package.json
-    // and it's not necessarily `index.mjs`, might be index.cjs, app.mjs, etc.
     const packageJsonContent = JSON.parse(contents["package.json"]) as PackageJSON;
-    return packageJsonContent.main || "index.mjs";
+    const mainPath = packageJsonContent.main;
```
costinsin, line 109: not sure if we want to support this, but there are multiple unofficial entrypoints in package.json: https://esbuild.github.io/api/#main-fields

costinsin (approved):
LGTM

src/commands/analyze/constants.ts
andreia-oca: This is seems to be the object/import on the official documentation:
```
// ESM
import Fastify from 'fastify'

const fastify = Fastify({
  logger: true
})
```

https://www.npmjs.com/package/fastify

src/commands/analyze/constants.ts
andreia-oca: I guess this is the same - but here I am not 100% sure.
https://github.com/Genez-io/express-getting-started/blob/main/index.mjs#L2

src/commands/analyze/frameworks.ts
andreia-oca: I thought about it and I'm not gonna over-engineer this.
Looking for the main entry in `package.json` it's just an heuristic to decide faster what is the main entry file.

We have a fallback to search for the main file recursively.

Thanks for suggesting it, tho' - TIL.

src/commands/analyze/frameworks.ts
andreia-oca: Great suggestion - thanks

PR #1504, opened 2024-11-11
Detect genezio typesafe

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [X] 🧑‍💻 Improvement

## Description

Detect Genezio Typesafe classes as best as we can.
We are taking into account only the decorators way (not classes defined in yaml) and usually the code is biased towards typescript.

## Checklist

-   [ ] I have added tests;


cristim67 (approved):
LGTM

src/commands/analyze/command.ts
```diff
@@ -486,6 +506,7 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
     // Inject Backend API URLs into the frontend component
     // This is done after all the components have been detected
     if (
+        !frameworksDetected.backend?.includes("genezio-typesafe") &&
```
vladiulianbogdan, line 509: I think if you reverse these conditions you won't need `?` anymore. Also, it would be more natural to read.

```
 if (
        frameworksDetected.backend &&
        frameworksDetected.backend.length > 0 &&
        !frameworksDetected.backend.includes("genezio-typesafe") &&
        frameworksDetected.frontend &&
```

src/commands/analyze/utils.ts
```diff
@@ -146,6 +147,49 @@ export async function injectBackendApiUrlsInConfig(configPath: string, frontendP
...
+ * @param configPath The path to the config file.
+ */
+export async function injectSDKInConfig(configPath: string) {
+    const configIOController = new YamlConfigurationIOController(configPath);
+
+    // Load config with minimal changes
+    const config = await configIOController.read(/* fillDefaults= */ false);
+
+    const frontend = config.frontend as YamlFrontend;
+    if (!frontend) return;
```
vladiulianbogdan, line 161: I would have expected this to be in an `if` before calling this method.

andreia-oca: I moved the check - nice catch

PR #1555, opened 2024-12-02
Reimplement the entryfile detection algorithm

andreia-oca:
Now, we are searching the package.json for a main module. If this is found in the filesystem, we are setting that as an entryfile. If the user defined and used the main file as set in package.json, we will trust he set it correctly.

Else, we are searching for patterns and set it as an entryfile.

This approach is less error-prone.

# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

Improve the way we detect the entryfile for servers - 
1. We search for a `package.json` - if the main file is set and it exists on the file system => we set it as an entryfile
2. Else, we search for files with corresponding patterns => if we find, we set it as an entryfile.
3. If nothing matched from abode, we return a default.

Use-case: [This repository](https://github.com/socketio/chat-example/blob/main/index.js#L81) has an entryfile correctly set, but it doesn't match our patterns (uses `server.listen` instead of `app.listen`)

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/analyze/frameworks.ts
+++ b/src/commands/analyze/frameworks.ts
@@ -2,6 +2,7 @@ import { promises as fs } from "fs";
 import path from "path";
 import { EXCLUDED_DIRECTORIES, KEY_DEPENDENCY_FILES } from "./command.js";
 import { FUNCTION_EXTENSIONS } from "../../models/projectOptions.js";
+import { debugLogger } from "../../utils/logging.js";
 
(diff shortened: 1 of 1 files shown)


cristim67 (approved):
LGTM

PR #1591, opened 2024-12-10
test interface not available if no classes found

Virgil993:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

Context: we want to push the users to test their functions from postman instead of our test interface because our test interface is not as robust and effective as postman. As such, the new version of the test interface will handle functions simillar to how we handle http methods and alert the user to use Postman instead. If the user doesn't have any classes then the test interface will not be available all togheter. It will be shown only if the user has a minimum of one class

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -1416,11 +1416,13 @@ function reportSuccess(projectConfiguration: ProjectConfiguration, port: number)
     }
 
     const workspaceUrl = getWorkspaceUrl(port);
-    log.info(
-        colors.cyan(
-            `Test your code at ${workspaceUrl ? workspaceUrl : `http://localhost:${port}`}/explore`,
-        ),
-    );
+    if (projectConfiguration.classes.length > 0) {
+        log.info(
+            colors.cyan(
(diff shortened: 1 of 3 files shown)


src/commands/local.ts
```diff
@@ -1416,11 +1416,13 @@ function reportSuccess(projectConfiguration: ProjectConfiguration, port: number)
...
     const workspaceUrl = getWorkspaceUrl(port);
-    log.info(
-        colors.cyan(
-            `Test your code at ${workspaceUrl ? workspaceUrl : `http://localhost:${port}`}/explore`,
-        ),
-    );
+    if (projectConfiguration.classes.length > 0) {
+        log.info(
+            colors.cyan(
+                `Test your code at ${workspaceUrl ? workspaceUrl : `http://localhost:${port}`}/explore`,
```
andreia-oca, line 1422:
```suggestion
                `Test your classes at ${workspaceUrl ? workspaceUrl : `http://localhost:${port}`}/explore`,
```

andreia-oca (approved):

PR #1594, opened 2024-12-10
[BREAKING CHANGE] Enhance genezio analyze env vars

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

This PR makes an improvement on reporting the environment variables to the dashboard.
Now every set of environment variables is mapped to the corresponding component.
The mapping is done using the `path` which must be the same for the component and for the `.env.example` file.

```
export type FrameworkReport = {
    backend?: BaseComponent[];
    frontend?: BaseComponent[];
    ssr?: BaseComponent[];
    services?: FrameworkReportService[];
};

export type BaseComponent = {
    component: string;
    path?: string;
    environment?: ProjectEnvironment[];
};
```

This is a breaking change for the frontend because it updates the objects passed to the dashboard - this PR should be merged with it's companion on the frontend side (wip).

The environment variables are retrieved and set using an OpenAI agent, if that's not available (no OpenAI key found), the fallback is to parse and load each `.env.example` as it is (with keys and potentially default values).

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/analyze/command.ts
+++ b/src/commands/analyze/command.ts
@@ -261,6 +261,9 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
                 {
                     path: componentPath,
                     packageManager: getPackageManager().command as PackageManagerType,
+                    environment: mapEnvironmentVariableToConfig(
+                        resultEnvironmentAnalysis.get(componentPath)?.environmentVariables,
+                    ),
                     scripts: {
                         deploy: [`${getPackageManager().command} install`],
                     },
@@ -268,7 +271,10 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
                 SSRFrameworkComponentType.nestjs,
             );
             frameworksDetected.ssr = frameworksDetected.ssr || [];
-            frameworksDetected.ssr.push("nest");
+            frameworksDetected.ssr.push({
+                component: "nestjs",
+                environment: resultEnvironmentAnalysis.get(componentPath)?.environmentVariables,
+            });
             continue;
         }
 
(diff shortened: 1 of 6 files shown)


src/commands/analyze/command.ts
```diff
@@ -97,10 +99,23 @@ export enum SUPPORTED_FORMATS {
 export const DEFAULT_FORMAT = SUPPORTED_FORMATS.TEXT;
 export const DEFAULT_CI_FORMAT = SUPPORTED_FORMATS.JSON;
 
-export const KEY_FILES = ["package.json", "requirements.txt", "pyproject.toml"];
 export const KEY_DEPENDENCY_FILES = ["package.json", "requirements.txt", "pyproject.toml"];
 export const ENVIRONMENT_EXAMPLE_FILES = [".env.template", ".env.example", ".env.local.example"];
```
cristim67, line 103: can you also add `.env.sample`

src/commands/analyze/command.ts
```diff
@@ -637,6 +710,71 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
...
 
+async function analyzeEnvironmentFilesConcurrently(
+    envExampleFiles: Map<string, string>,
+    genezioConfig: RawYamlProjectConfiguration,
+): Promise<
+    Map<
+        string,
+        {
+            environmentVariables: ProjectEnvironment[];
+        }
```
cristim67, line 721: The return type `Map<string, { environmentVariables: ProjectEnvironment[] }> ` is repeated multiple times. Define it as a separate type for maintainability:

```
type EnvironmentAnalysisResult = Map<string, { environmentVariables: ProjectEnvironment[] }>;
```

src/commands/analyze/command.ts
```diff
@@ -637,6 +710,71 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
     log.info(result);
 }
 
+async function analyzeEnvironmentFilesConcurrently(
```
cristim67, line 713: Handle this edge case:

```
  if (envExampleFiles.size === 0) {
        return new Map();
    }
    
 ```

cristim67 (approved):

src/commands/analyze/command.ts
andreia-oca: Nice catch <3

PR #1602, opened 2024-12-16
[DRAFT] SSR - genezio local

cristim67:
## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

Add SSRs in genezio local

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -124,6 +124,14 @@ type PortMapping = {
 
 const httpServerPortMapping: PortMapping = {};
 
+type SsrFramework = {
+    path: string;
+    scripts?: {
+        start?: string;
+    };
+    environment?: Record<string, string>;
+};
+
 export async function prepareLocalBackendEnvironment(
     yamlProjectConfiguration: YamlProjectConfiguration,
     options: GenezioLocalOptions,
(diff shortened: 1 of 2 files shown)


src/commands/local.ts
```diff
@@ -329,7 +330,14 @@ export async function startLocalEnvironment(options: GenezioLocalOptions) {
...
     sdkSynchronizer.acquire();
 
-    if (!yamlProjectConfiguration.backend && !yamlProjectConfiguration.frontend) {
+    if (
+        !yamlProjectConfiguration.backend &&
+        !yamlProjectConfiguration.frontend &&
+        !yamlProjectConfiguration.nextjs &&
+        !yamlProjectConfiguration.nuxt &&
+        !yamlProjectConfiguration.nestjs &&
+        !yamlProjectConfiguration.nitro
```
andreia-oca, line 339: Rebase this PR and add `remix` too.

src/commands/local.ts
```diff
@@ -1751,3 +1776,118 @@ else:
...
+            input: childProcess.stderr!,
+        });
+
+        const stripAnsi = (str: string): string => {
+            /* eslint-disable-next-line no-control-regex */
+            return str.replace(
+                /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
+                "",
+            );
+        };
```
andreia-oca, line 1859: This looks ugly :)) Why are we attempting to remove something from the output?
Can we maybe control this from `npx` itself?

cristim67: some framework like next, clear the terminal output when starts

src/commands/local.ts
```diff
@@ -1751,3 +1776,118 @@ else:
...
+            port: port,
+        },
+    );
+
+    debugLogger.debug(
+        `Environment variables injected for ${frameworkName}:`,
+        JSON.stringify(newEnvObject),
+    );
+
+    const ssrPort = await findAvailablePort();
```
andreia-oca, line 1806: nit - Usually this is not a random port, it's 3000, and if that is not available it tries every port until it finds one (3001, 3002) and so on.
I would do the same because next users are used to test directly on `localhost:3000` most of the time.

andreia-oca (approved):

src/commands/local.ts
cristim67: I will come back later with an pr for this.

PR #1608, opened 2024-12-18
Handle Nextjs fullstack applications

cristim67:
## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Handle Nextjs fullstack applications

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/nextjs/deploy.ts
+++ b/src/commands/deploy/nextjs/deploy.ts
@@ -46,13 +46,12 @@ import {
 import { readOrAskConfig } from "../utils.js";
 import { SSRFrameworkComponentType } from "../../../models/projectOptions.js";
 import { addSSRComponentToConfig } from "../../analyze/utils.js";
-
 export async function nextJsDeploy(options: GenezioDeployOptions) {
     const genezioConfig = await readOrAskConfig(options.config);
     const packageManagerType = genezioConfig.nextjs?.packageManager || NODE_DEFAULT_PACKAGE_MANAGER;
 
     let cwd = process.cwd();
-    const componentPath = genezioConfig.nextjs?.path
+    let componentPath = genezioConfig.nextjs?.path
         ? path.resolve(cwd, genezioConfig.nextjs.path)
         : cwd;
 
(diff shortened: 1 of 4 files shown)


src/commands/deploy/nextjs/deploy.ts
```diff
@@ -409,40 +423,78 @@ function writeNextConfig(cwd: string, region: string) {
...
 
-    const configPath = path.join(cwd, `next.config.${existingConfig}`);
-    const handlerPath = `./cache-handler.${existingConfig}`;
+    const currentConfigPath = path.join(cwd, `next.config.${existingConfig}`);
+    const baseNextConfigPath = path.join(cwd, `base-next.${existingConfig}`);
+
+    // Rename next.config.{ext} to base-next.{ext}
+    fs.renameSync(currentConfigPath, baseNextConfigPath);
+
+    const genezioConfigPath = path.join(cwd, `next.config.${existingConfig}`);
```
andreia-oca, line 434: This seems to be the same with `currentConfigPath` - maybe instead of `currentConfigPath` and `baseNextConfigPath` use `genezioConfigPath` and `userConfigPath`.

src/commands/deploy/nextjs/deploy.ts
```diff
@@ -46,13 +46,12 @@ import {
...
 import { addSSRComponentToConfig } from "../../analyze/utils.js";
-
 export async function nextJsDeploy(options: GenezioDeployOptions) {
     const genezioConfig = await readOrAskConfig(options.config);
     const packageManagerType = genezioConfig.nextjs?.packageManager || NODE_DEFAULT_PACKAGE_MANAGER;
 
-    const cwd = process.cwd();
-    const componentPath = genezioConfig.nextjs?.path
+    let cwd = process.cwd();
+    let componentPath = genezioConfig.nextjs?.path
```
andreia-oca, line 54: These should not be defined with `let` because it will become confusing - we will start using this `cwd` in-between declarations and it will not be clear when it's `process.cwd` and when is not.

IMO:
- `cwd` should be where `genezio.yaml` is and where `genezio deploy` was ran, 
- `componentPath` is the relative path between `cwd` and the nextjs folder - the componentPath is saved in yaml at `next.path` and it's where we are running scripts and commands such as `npm install`.

For the `temp` folder processing and building I think you should use new vars named  something like`tempBuildCwd` and `tempBuildComponentPath`.

src/commands/deploy/nextjs/deploy.ts
```diff
@@ -64,60 +64,65 @@ export async function nextJsDeploy(options: GenezioDeployOptions) {
...
     // Add nextjs component
     await addSSRComponentToConfig(
         options.config,
         {
-            path: componentPath,
+            path: nextjsComponentPath,
             packageManager: packageManagerType,
-            scripts: {
-                deploy: [`${installDependenciesCommand.command}`],
-            },
```
andreia-oca, line 89: Nit - We put this in the `genezio.yaml` to allow the user to know he can control the install script and add `--legacy-peer-deps`. You can move the whole `addSSRComponentToConfig` chunk after installing deps below to keep this.

andreia-oca (approved):
LGTM

src/commands/deploy/nextjs/deploy.ts
cristim67: I don't necessarily like it; I removed it because we handle the installation ourselves.
If there's an npm install in the YAML, two install commands will run: the one in the YAML as a pre-command, followed by what we actually do.

```
scripts:
    deploy:
        - npm install

```

PR #1641, opened 2025-01-10
GNZ-2183 Add start script for frontend components

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

Adding start scripts for the frontend components will enable `genezio local` to start the project in a fullstack manner.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/analyze/command.ts
+++ b/src/commands/analyze/command.ts
@@ -470,6 +470,10 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
                 scripts: {
                     deploy: [`${packageManager.command} install`],
                     build: [`${packageManager.command} run build`],
+                    start: [
+                        `${packageManager.command} install`,
+                        `${packageManager.command} run dev`,
+                    ],
                 },
             });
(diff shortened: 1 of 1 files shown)


src/commands/analyze/command.ts
```diff
@@ -470,6 +470,10 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
                 scripts: {
                     deploy: [`${packageManager.command} install`],
                     build: [`${packageManager.command} run build`],
+                    start: [
+                        `${packageManager.command} install`,
+                        `${packageManager.command} run dev`,
+                    ],
```
cristim67, line 476: personally, I don't like that every time I run `genezio local`, I will run `npm install` in this approach.

for me, the logical line would be to have the dependencies already installed and then to run genezio local. otherwise I get an basic error.

we can check in the genezio local command if it has a node_modules and run the install command if it doesn't.

PR #1643, opened 2025-01-10
Add support Remix.js

cristim67:
## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

# Add Remix Framework Deployment Support

## Technical Overview
This PR implements Remix deployment support, handling both Remix Vite and Classic Remix architectures.

### Build System Detection & Handling
The implementation dynamically identifies and manages two distinct build patterns:

1. **Remix Vite Build**
   - Uses the modern Vite bundler
   - Output structure:
     ```
     build/
     ├── client/        # Client-side assets
     └── server/        # Server-side bundle
         └── index.js   # Server entry point
     ```
 
2. **Classic Remix Build**
   - Traditional Remix bundler
   - Output structure:
     ```
     build/
     ├── index.js      # Server entry point
     └── public/       # Static assets
     ```

In both cases index.js will always be a module bundle.

### Server Implementation

Both variants are served through an Express.js server with specific configurations:

```
import express from "express";
import { createRequestHandler } from "@remix-run/express";
import * as build from "./index.js";
const app = express();

app.use(express.static('public'));
app.all('*', createRequestHandler({ build }));

app.listen(8080, () => {
  console.log('Server running at http://localhost:8080');
});
```

### Example 

https://d68cc551-3b9b-4d08-b0cd-50fa15300aa9.dev-fkt.cloud.genez.io/

### References

https://remix.run/docs/en/main/start/quickstart#bring-your-own-server

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/analyze/command.ts
+++ b/src/commands/analyze/command.ts
@@ -550,6 +550,36 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
         }
 
         if (await isViteComponent(contents)) {
+            if (await isRemixComponent(contents)) {
+                const packageManagerType =
+                    genezioConfig.remix?.packageManager || NODE_DEFAULT_PACKAGE_MANAGER;
+                const packageManager = packageManagers[packageManagerType];
+                await addSSRComponentToConfig(
+                    options.config,
+                    {
+                        path: componentPath,
+                        packageManager: packageManagerType,
+                        environment: mapEnvironmentVariableToConfig(
+                            resultEnvironmentAnalysis.get(componentPath)?.environmentVariables,
+                        ),
+                        scripts: {
+                            build: [`${packageManager.command} run build`],
+                            deploy: [
+                                `${packageManager.command} install`,
+                                `${packageManager.command} run build`,
+                            ],
+                        },
+                    },
+                    SSRFrameworkComponentType.remix,
+                );
+                frameworksDetected.ssr = frameworksDetected.ssr || [];
+                frameworksDetected.ssr.push({
+                    component: "remix",
+                    environment: resultEnvironmentAnalysis.get(componentPath)?.environmentVariables,
+                });
+                continue;
+            }
+
             const packageManagerType = NODE_DEFAULT_PACKAGE_MANAGER;
             const packageManager = packageManagers[packageManagerType];
             await addFrontendComponentToConfig(configPath, {
--- a/src/commands/analyze/command.ts
+++ b/src/commands/analyze/command.ts
@@ -23,6 +23,7 @@ import {
     hasPostgresDependency,
     hasMongoDependency,
     isNestjsComponent,
+    isRemixComponent,
 } from "./frameworks.js";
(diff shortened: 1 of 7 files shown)


src/commands/analyze/command.ts
```diff
@@ -519,6 +550,36 @@ export async function analyzeCommand(options: GenezioAnalyzeOptions) {
         }
 
         if (await isViteComponent(contents)) {
+            if (await isRemixComponent(contents)) {
```
andreia-oca, line 553: I believe this code will never be reached - because there is an if `if (await isRemixComponent(contents))` above that will trigger a continue.

src/commands/deploy/command.ts
```diff
@@ -181,6 +189,13 @@ async function decideDeployType(options: GenezioDeployOptions): Promise<DeployTy
         ) {
             return [DeployType.Nest];
         }
+        if (
+            Object.keys({ ...packageJson.dependencies, ...packageJson.devDependencies }).some(
+                (dep) => dep.startsWith("@remix-run/"),
```
andreia-oca, line 194:
```suggestion
                (dep) => dep.startsWith("@remix-run"),
```

src/commands/deploy/remix/deploy.ts
```diff
@@ -0,0 +1,280 @@
...
+        // Build the project using remix build
+        await $({
+            stdio: "inherit",
+            cwd: componentPath,
+        })`npx remix build`.catch(() => {
+            throw new UserError("Failed to build the Remix project. Check the logs above.");
+        });
+    }
+
+    const remixBuildCommand = isRemixVite ? "remix vite:build" : "npx remix build";
```
andreia-oca, line 80: Here we are assuming that `remix` is installed on the user's machine. Is that an okay assumption?

src/commands/deploy/remix/deploy.ts
```diff
@@ -0,0 +1,280 @@
...
+            throw new UserError("Failed to copy project build folder to temporary directory");
+        });
+
+    if (!isRemixVite) {
+        // Copy public folder
+        const publicPath = path.join(componentPath, "public");
+        await fs.promises
+            .mkdir(path.join(tempBuildCwd, "public"), { recursive: true })
+            .catch(() => {
+                throw new UserError("Failed to create public folder in temporary directory");
```
andreia-oca, line 114: We are not displaying the root cause error. It's hard as a user to understand what I can do actionable after receiving this error.

andreia-oca (approved):
LGTM

src/commands/deploy/remix/deploy.ts
cristim67: Yes, for `remix build` I install the dependencies above, and for `npx remix build`, the npx command will prompt the user for installation.

`npx remix build` will be use for legacy projects.

PR #1651, opened 2025-01-14
Allow Users to Set the Port for SSR Frameworks

cristim67:
## Type of change

-   [ ] 🍕 New feature
-   [x] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

### Allow Users to Set the Port for SSR Frameworks

Updated logic to let users configure the port via an environment variable for frameworks like Nest.js. If the port is not explicitly set, it falls back to a default. For Nest.js, a user-friendly error is thrown if the port is missing.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/local.ts
+++ b/src/commands/local.ts
@@ -1857,15 +1857,15 @@ async function startSsrFramework(
         switch (frameworkName.toLowerCase()) {
             case "next.js":
                 command = "next";
-                args = ["dev", "--port", ssrPort.toString()];
+                args = ["dev", "--port", process.env[portEnvKey]];
                 break;
             case "nuxt":
                 command = "nuxt";
-                args = ["dev", "--port", ssrPort.toString()];
+                args = ["dev", "--port", process.env[portEnvKey]];
                 break;
             case "nitro":
                 command = "nitropack";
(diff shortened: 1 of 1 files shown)


src/commands/local.ts
```diff
@@ -1829,14 +1829,20 @@ async function startSsrFramework(
...
+2. Running \`set GENEZIO_PORT_NEST_JS=<port> && genezio local\` - for windows
+3. Adding \`GENEZIO_PORT_NEST_JS=<port>\` to your \`.env\` file.
+            `,
+        );
     }
 
-    process.env[`GENEZIO_PORT_${frameworkName.replace(/[^a-zA-Z0-9]/g, "_").toUpperCase()}`] =
-        ssrPort.toString();
+    const ssrPort = process.env[portEnvKey] || (await findAvailablePort()).toString();
+    process.env[portEnvKey] = ssrPort;
```
andreia-oca, line 1844: You don't need this line anymore, right?

If I'm wrong, disregard

andreia-oca (approved):
Looks great 🎉

PR #1680, opened 2025-01-27
feat: add frontend name filter for selective deployment

cristim67:
## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

This change allows users to deploy specific frontends by name using the --name flag. 
When provided, only the frontend matching the specified name will be deployed. 
If no matching frontend is found, the command will fail with a clear error message.

Example usage:
genezio deploy --frontend --name my-frontend

## Checklist

-   [ ] I have added tests;


src/genezio.ts
```diff
@@ -179,6 +179,7 @@ program
     .command("deploy")
     .option("--backend", "Deploy only the backend application.", false)
     .option("--frontend", "Deploy only the frontend application.", false)
+    .option("--name <name>", "Name of the component to deploy.")
```
andreia-oca, line 182:
```suggestion
    .option("--name <name>", "Specify the name of the frontend to be deployed .")
```

andreia-oca: This option doesn't make sense for deploying backend, ssr or containers. We should mildly specify that so the user won't be confused.

src/commands/deploy/genezio.ts
```diff
@@ -238,6 +243,11 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
                 commandOptions: JSON.stringify(options),
             });
         }
+
+        // If name was provided but no matching frontend was found, throw an error
+        if (options.name && frontendUrls.length === 0) {
```
andreia-oca, line 248: This check is a bit misleading - it doesn't do what the comment says.

It just checks if there are any frontendUrls deployed.

src/commands/deploy/genezio.ts
```diff
@@ -173,6 +173,11 @@ export async function genezioDeploy(options: GenezioDeployOptions) {
         const frontends = configuration.frontend;
 
         for (const [index, frontend] of frontends.entries()) {
+            // Skip if name option is provided and doesn't match the current frontend
+            if (options.name && frontend.name !== options.name) {
```
andreia-oca, line 177: `frontend.name` might be undefined because `name` is an optional field in the yaml.

A better UX would be if `options.name` is used, the user should be guided (via a warning log) to specify `frontend.name` for his frontends.

Otherwise - no frontend will be deployed because this condition will probably be false for every frontend.

src/commands/deploy/genezio.ts
andreia-oca: And imo the check is done pretty late - after iterating through all the frontends and attempting to deploy.

I think this check can be done outside the `for` loop - if `options.name` doesn't match any frontend.name in the yaml - we should give a warning that the `options.name` doesn't exists + check frontend.name and add them.

andreia-oca (changes requested):

andreia-oca (approved):
LGTM

PR #1690, opened 2025-01-30
Display deployment auth tokens on success

MagoDelBlocco:
-   [X] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

Newer deployments return an authentication token needed to query for individual logs, directly from our runtime, through a yet undocumented API. This token should be shown at deployment in the CLI.

This PR makes sure to capture the auth token from the backend, and then print it alongside the individual URLs of the deployment.

### DIFF
--- a/src/utils/reporter.ts
+++ b/src/utils/reporter.ts
@@ -27,6 +27,11 @@ export function reportSuccess(classesInfo: GenezioCloudResultClass[]) {
                     "\n";
             }
         });
+        if (classInfo.authToken) {
+            printHttpString +=
+                `  - ${classInfo.className} authentication token: ${colors.green(classInfo.authToken)}` +
+                "\n";
+        }
     });
 
     if (printHttpString !== "") {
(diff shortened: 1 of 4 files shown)


src/utils/reporter.ts
```diff
@@ -41,11 +46,15 @@ export function reportSuccessFunctions(functions: DeployCodeFunctionResponse[])
 
     functions.forEach((func) => {
         functionDeploymentsString += `  - ${func.name}: ${colors.yellow(func.cloudUrl)}` + "\n";
+        if (func.authToken) {
+            functionDeploymentsString +=
+                `  - ${func.name} authentication token: ${colors.green(func.authToken)}` + "\n";
```
andreia-oca, line 51: I am not a fan of revealing this information to the command line without a specific request from the user. Because it's sensitive info - if I am deploying in a public repository in GitHub Actions, this deployment secret id will be leaked.

I advise to either add a new command  something like `genezio reveal <function_cloud_id>` that would print secret informatio or to make it a frontend feature where it should have a reveal button.

andreia-oca: Also, user facing I would call this token - `deployment secret id`, it seems to me it's more descriptive.

MagoDelBlocco:
Opted for the alternative to show it in the frontend dashboard.

PR #1702, opened 2025-02-06
Prevent sending a runtime default to the backend

andreia-oca:
# Pull Request Template

## Type of change

-   [ ] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [x] 🧑‍💻 Improvement

## Description

To allow the backend to handle backwards compatibility issues related to latest `language.runtime` we should prevent pushing `nodeRuntime` and `pythonRuntime` to the backend. These should be set only when the user specifically sets them in the `genezio.yaml`.

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/deploy/nestjs/deploy.ts
+++ b/src/commands/deploy/nestjs/deploy.ts
@@ -19,7 +19,6 @@ import {
 } from "../../../packageManagers/packageManager.js";
 import {
     DEFAULT_ARCHITECTURE,
-    DEFAULT_NODE_RUNTIME,
     SSRFrameworkComponentType,
 } from "../../../models/projectOptions.js";
 import { UserError } from "../../../errors.js";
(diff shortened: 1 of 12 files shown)


src/commands/deploy/nestjs/deploy.ts
```diff
@@ -133,8 +137,8 @@ async function deployFunction(
             path: cwdRelative,
             language: {
                 name: Language.js,
-                runtime: "nodejs20.x",
-                architecture: "x86_64",
+                runtime: DEFAULT_NODE_RUNTIME,
+                architecture: DEFAULT_ARCHITECTURE,
```
cristim67, line 141: What happens if I already set the desired version in the YAML file and it's different from the default? 

```suggestion
                runtime: (config.nestjs?.runtime as NodeRuntime) || DEFAULT_NODE_RUNTIME,
                architecture: (config.nestjs?.architecture as Architecture) || DEFAULT_ARCHITECTURE,
```

same suggestion for the rest of the ssr type

cristim67: And one more thing, for this suggestion, you should extend the SSR schema with the architecture type.

cristim67 (approved):

src/commands/deploy/nestjs/deploy.ts
andreia-oca: `architecture` is not yet a field that's used, I don't want to make changes to promote it to our users.

PR #1767, opened 2025-03-21
Add a command to get environment variables

andreia-oca:
# Pull Request Template

## Type of change

-   [x] 🍕 New feature
-   [ ] 🐛 Bug Fix
-   [ ] 🔥 Breaking change
-   [ ] 🧑‍💻 Improvement

## Description

This PR introduces `genezio getenv` - a useful command to reveal env vars values.

This command doesn't require a `genezio.yaml` to work. It can be used as following:
```
genezio getenv --projectName <name>
genezio getenv --projectName <name> --format env > .env
genezio getenv --projectName <name> --format json | jq
GENEZIO_TOKEN=<token> genezio getenv --projectName
```

## Checklist

-   [ ] I have added tests;

### DIFF
--- a/src/commands/env.ts
+++ b/src/commands/env.ts
@@ -3,10 +3,16 @@ import { GenezioEnvOptions } from "../models/commandOptions.js";
 import { revealEnvironmentVariablesRequest } from "../requests/getEnvironmentVariables.js";
 import getProjectInfoByName from "../requests/getProjectInfoByName.js";
 import { log } from "../utils/logging.js";
+import { writeToFile } from "../utils/file.js";
 
 export async function revealEnvironmentVariables(options: GenezioEnvOptions) {
     const projectName = options.projectName;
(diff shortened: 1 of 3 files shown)


src/commands/env.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const envVarList = await revealEnvironmentVariablesRequest(project.id, environment.id);
+
+    if (options.format == "json") {
+        log.info(JSON.stringify(envVarList));
+        return;
+    }
+
+    for (const envVar of envVarList) {
+        log.info(`${envVar.name}=${envVar.value}`);
```
vladiulianbogdan, line 31: I think it is better if we don't print the env vars to stdout. Let's write them in a file that is passed as parameter to the CLI command.

andreia-oca: Done

